PR #2, opened 2019-12-13
Added webpack-dev-server for quick development process

logvik:
Added a logic for the encrypt / decrypt message feature testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/node_modules/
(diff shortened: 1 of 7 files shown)


whymarrh:
@logvik thanks for putting the work into this originally & sorry we didn't get to it.

I'm going to close this for now as `master` has drifted quite a bit.

whymarrh:
We have #20 open to make sure this feature is represented in the app

PR #4, opened 2020-01-10
Prevent duplicate click listeners, prepare for gh-pages

rekmarks:
- Copy latest test dapp from `metamask-extension#develop`
- Add `@metamask/onboarding` dependency
- Prevent duplicate click listeners: `s/addEventListener('click', /onclick(`
- Prepare for `gh-pages` deployment
  - add deploy script that pushes relevant files to `gh-pages` branch
  - use `unpkg.com/metamask-onboarding` as source for onboarding dependency in `gh-pages`


whymarrh:
How does this work? The [`onlick`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onclick) can b a function but is it assigned anywhere?

rekmarks:
> How does this work? The [`onlick`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onclick) can b a function but is it assigned anywhere?

@whymarrh I'm not sure what you mean. We're just replacing `addEventListener('click', callback)` with `onclick(callback)`. The former adds a new event listener for the given event on the target object; the latter sets a click listener, replacing the existing one, if any.

This is to address a bug in e2e tests caused by duplicate listeners.

Note that these changes are very outdated, and are about to be replaced.

deploy.sh
```diff
@@ -0,0 +1,110 @@
...
+  sed -i "" -e "s/${local_src}/${web_src}/g" "${target_file}"
+}
+
+function main {
+
+  local gh_remote='origin'
+  local source_branch='master'
+  local deploy_branch='gh-pages'
+
+	while :; do
```
Gudahtt: This file appears to have a mix of tabs and spaces 🤔

rekmarks: Replaced with spaces

contract.js
```diff
@@ -41,154 +47,308 @@ web3.currentProvider.enable().then(() => {
...
+            }, (result) => {
               console.log(result)
             })
-          })
+          }
+        },
+      )
+    }
+
+    signTypedData.onclick(() => {
```
Gudahtt:
```suggestion
    signTypedData.onclick = () => {
```
Shouldn't this be assigned the function? I'm not sure what's going on here by calling `onclick`, since it hasn't been given a value yet.

Gudahtt: I think this is what @whymarrh meant. See the other `onclick` listeners for examples of this.

rekmarks: Aha, got it. Fixed!

deploy.sh
```diff
@@ -0,0 +1,111 @@
...
+    shift
+  done
+
+  if is_working_tree_dirty
+  then
+    die 'Working tree is dirty; please clean it up and try again'
+  fi
+
+  # checkout gh-pages, update files
+  if [ `git branch --list $deploy_branch`]
```
Gudahtt: [shellcheck](https://www.shellcheck.net/) didn't seem to like this line

For one thing, I'd suggest using `[[ ... ]]` instead of `[ ... ]` as it behaves a little more sanely.

Also, `git branch --list` is meant for humans. This is probably better achieved with `git show-ref "refs/heads/${deploy_branch}"`

And actually, with `git show-ref` you wouldn't need the test command at all. That command returns a value that evaluates as true if it exists, false otherwise, so.... this should work:
```suggestion
  if git show-ref "refs/heads/${deploy_branch}"
```

deploy.sh
```diff
@@ -0,0 +1,111 @@
...
+  then
+    git checkout "${deploy_branch}"
+  else
+    git checkout -b "${deploy_branch}"
+    rm *
```
Gudahtt: Hmm, this seems a bit dangerous 🤔 . Shellcheck has also pointed out that it doesn't protect against files that start with a dash being parsed as arguments.

Maybe something like this would be better:
```suggestion
    git rm -r ./*
```

deploy.sh
```diff
@@ -0,0 +1,111 @@
...
+  fi
+  git checkout "${source_branch}" -- contract.js index.html || die "Failed to checkout files from ${source_branch}"
+
+  # make changes for web publication
+  replace_onboarding_src
```
Gudahtt: Note that I didn't use `set -e` in the script that this is derived from, instead opting to explicitly check for errors after every command that might fail. This command for instance should be followed by an `|| die "error message"`. Either that or you could turn on `set -e`, following the pattern we use in most of the `metamask-extension` Bash scripts.

deploy.sh
```diff
@@ -0,0 +1,111 @@
...
+
+  git commit -am "update to latest master" || die "Failed to commit to ${deploy_branch}"
```
Gudahtt: It would be helpful in this commit message to include the short hash of the commit on `master` that triggered this one

deploy.sh
```diff
@@ -0,0 +1,111 @@
...
+  git push -f -u ${gh_remote} ${deploy_branch} || die "Failed to push to ${gh_remote}/${deploy_branch}"
```
Gudahtt: I think the `-u` might be unnecessary? 🤔 . Not 100% sure, but I thought that was just to setup the local branch to track the remote one, which isn't a useful thing to do here.

The `-f` also seems unnecessary, as it doesn't look like there's any need for rewriting the git history of the deploy branch,

rekmarks: Fair enough on the `-f`.

Updated line to: https://github.com/MetaMask/test-dapp/pull/4/files#diff-60254338249f657a0a83f98258a56bfeR56

Following: https://stackoverflow.com/questions/1783405/how-do-i-check-out-a-remote-git-branch#1783426

In case the branch doesn't exist locally. Also included a `git fetch`.

deploy.sh
rekmarks: Fixed in latest

deploy.sh
```diff
@@ -0,0 +1,134 @@
...
+  local local_src="node_modules/@metamask/onboarding/dist/metamask-onboarding.bundle.js"
+  local web_src="unpkg.com/metamask-onboarding"
+  local target_file="index.html"
+
+  sed -i "" -e "s/${local_src}/${web_src}/g" "${target_file}"
+}
+
+function preprocess_and_publish {
+
+  git fetch
```
Gudahtt: This, and most of the other git commands, should have error handling as well

deploy.sh
```diff
@@ -0,0 +1,134 @@
...
+
+  # checkout gh-pages, update local files
+  if git show-ref "refs/heads/${DEPLOY_BRANCH}"
+  then
+    git checkout "${DEPLOY_BRANCH}"
+    git pull
```
Gudahtt:
```suggestion
    git reset --hard "${GH_REMOTE}/${DEPLOY_BRANCH}"
```
This would probably be better than `git pull`, as we wouldn't want to create unnecessary merge commits.

deploy.sh
```diff
@@ -0,0 +1,134 @@
...
+  then
+    abort "Source branch '${SOURCE_BRANCH}' has no head"
+  fi
+
+  _hash="${_hash:6}"
+
+  # commit to destination branch with shorthash in message
+  git commit -am "update using ${SOURCE_BRANCH}/${_hash}" || abort "Failed to commit to destination branch '${DEPLOY_BRANCH}'"
+
+  git push -u "${GH_REMOTE}" "${DEPLOY_BRANCH}" || abort "Failed to push to '${GH_REMOTE}/${DEPLOY_BRANCH}'"
```
Gudahtt: Was this `-u` left here intentionally? It should already be tracking that remote because of the `-t` above.

rekmarks: Unintentionally. Fixed!

deploy.sh
rekmarks: Added

Gudahtt:
I just updated the `metamask-extension` version of this test dapp slightly (https://github.com/MetaMask/metamask-extension/commit/3a2c81d8bca7fdcddbc2fb6b50e3fc51e7156f31) but I can update that here in a subsequent PR

PR #8, opened 2020-01-17
Fix test-dapp, add static-server

rekmarks:
Fixes `@metamask/onboarding` import in test dapp, updates readme, and adds `static-server` for local testing.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -4,23 +4,23 @@ set -u
 set -o pipefail
 
 readonly __SCRIPT_NAME__="${0##*/}"
-readonly __SEE_HELP_MESSAGE__="See '$__SCRIPT_NAME__ --help' for more information."
+readonly __SEE_HELP_MESSAGE__="See '${__SCRIPT_NAME__} --help' for more information."
 
 GH_REMOTE='origin'
 SOURCE_BRANCH='master'
 DEPLOY_BRANCH='gh-pages'
 
 function abort {
-  local message="$1"
+  local message="${1}"
 
-  printf "ERROR: %s\\n" "$message" >&2
+  printf "ERROR: %s\\n" "${message}" >&2
 
   exit 1
 }
 
 function show_help {
   cat << EOF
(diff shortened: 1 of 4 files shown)


deploy.sh
```diff
@@ -37,47 +37,49 @@ function is_working_tree_dirty {
 # for gh-pages, use unpkg.com for dependencies
 function replace_onboarding_src {
   local local_src="node_modules/@metamask/onboarding/dist/metamask-onboarding.bundle.js"
-  local web_src="unpkg.com/metamask-onboarding"
+  local web_src="https://unpkg.com/@metamask/onboarding"
   local target_file="index.html"
 
-  sed -i "" -e "s/${local_src}/${web_src}/g" "${target_file}"
+  sed -i "" -e "s#$local_src#$web_src#g" "$target_file"
```
Gudahtt: Was it necessary to remove all of the curly brackets? The unbound variable error you mentioned seeing on Slack may have been specific to the `@` variable - generally speaking the curly braces are good, as they remove the possibility of ambiguity (e.g. it might end up substituting a variable named `web` here instead of `web_src`, if one existed)

rekmarks: Ah, so they're not completely unnecessary? ;)

Added back

Gudahtt: Hah, I mean they're unnecessary for the case of `${@}`. There are no other characters in that string that could be confused as part of the variable name.

It's an old habit of mine that I don't see that often in the wild, and is admittedly a bit silly. I just started using braces for all variables at some point.

deploy.sh
```diff
@@ -107,7 +109,7 @@ function main {
         shift
         ;;
       -r|--remote)
-        if [[ -z $2 ]]; then
+        if [[ -z ${2} ]]; then
```
Gudahtt: I suspect this might fail. The curly braces syntax was meant for use within strings, and this isn't in a string. Variables don't need to be quoted in `[[` conditions because... it magically prevents string splitting somehow I guess.

PR #9, opened 2020-01-18
Update test dapp

rekmarks:
Fixes strange test dapp behavior. It should now be compatible with the old and new inpage providers. I've done basic manual testing with both.

In sum:
- Successfully retrieves Ethereum accounts with new and old inpage providers
- Updates accounts, networkId, and chainId always if possible
- `window.ethereum` -> `ethereum`
- Minor fixes to deploy script and package.json

### DIFF
--- a/contract.js
+++ b/contract.js
@@ -33,7 +33,7 @@ The `piggybankContract` is compiled from:
 const forwarderOrigin = 'http://localhost:9010'
 
 const isMetaMaskInstalled = () => {
-  return Boolean(window.ethereum && window.ethereum.isMetaMask)
+  return Boolean(ethereum && ethereum.isMetaMask)
 }
 
 const initialize = () => {
@@ -66,6 +66,7 @@ const initialize = () => {
   }
   let accounts
   let piggybankContract
+  let connected = false
 
   const accountButtons = [
     deployButton,
@@ -86,10 +87,6 @@ const initialize = () => {
     onboardButton.innerText = 'Onboarding in progress'
     onboardButton.disabled = true
     onboarding.startOnboarding()
-  }
-
-  const onClickConnect = async () => {
-    await window.ethereum.enable()
   }
 
   const updateButtons = () => {
@@ -309,7 +306,7 @@ const initialize = () => {
           contents: 'Hello, Bob!',
         },
       }
-      web3.currentProvider.sendAsync({
+      ethereum.sendAsync({
         method: 'eth_signTypedData_v3',
         params: [ethereum.selectedAddress, JSON.stringify(typedData)],
         from: ethereum.selectedAddress,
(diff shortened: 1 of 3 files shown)


whymarrh:
> Add latest Bootstrap CSS from their CDN for the benefit of mobile

Did this change get dropped?

rekmarks:
> Did [the bootstrap CSS] change get dropped?

Not quite, it seems I accidentally added that one in a fix to master, 3e7bb4c. So, it's already been done.

rekmarks:
@Gudahtt just pushed changes per your comments.

The `connected` boolean and calling `handleNewAccounts` on `ethereum.enable()` were, just as you pointed out, completely unnecessary. I replaced the former with calls to `isMetaMaskConnected` and removed the latter.

`getNetworkAndChainId`, on the other hand, is necessary in my local testing. `networkChanged`/`chainChanged` (and, the deprecated `chainIdChanged`) are emitted by the inpage provider on initialization _iff_ the subscription to the `publicConfigStore` state changes is registered before the `publicConfigStore` receives the network/chainId from the background. So, there is some uncertainty there.

In my local testing, I couldn't get the network/chainIds to populate properly without `getNetworkAndChainId`, and indeed, [EIP 1193](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#connect) suggests that the dapp gets the network/chainId after establishing that the provider is connected.

contract.js
```diff
@@ -66,6 +66,7 @@ const initialize = () => {
   }
   let accounts
   let piggybankContract
+  let accountButtonsInitialized = false
```
Gudahtt: This should be set to `false` again if the user disconnects from the dapp

Gudahtt:
>getNetworkAndChainId, on the other hand, is necessary per my local testing. networkChanged/chainChanged (and, the deprecated chainIdChanged) are emitted by the inpage provider on initialization iff the subscription to the publicConfigStore state changes is registered before the publicConfigStore receives the network/chainId from the background. So, there is some uncertainty there.

Huh, interesting, I didn't realize they were only populated sometimes. I'm just noticing now that they're populated about 1/3 of the time in local testing! Yeesh. I guess they just happened to get populated whenever I was paying attention 🤦‍♂ 

That behavior is concerning though 🤔 Ideally the provider would dispatch those events upon page load all the time, or none of the time. This makes our provider look kinda broken.

Reading that EIP, I'm not interpreting that as prescriptive - i.e. it's saying that you _can_ check the chainId after connecting, not that you _should_. But given how our provider is behaving today this seems like a good idea for the test dapp regardless.

contract.js
Gudahtt: Actually... maybe they don't need to be re-initialized ever? 🤔

Gudahtt: Yeah - while initializing them multiple times isn't harmful, it doesn't seem necessary either.

I'd go a step further and initialize them right away upon pageload, once, rather than waiting for the first connection and using this boolean to stop it the second time.

PR #11, opened 2020-01-21
Use shared MetaMask ESLint config

whymarrh:
This PR replaces the project's ESLint ruleset with the shared MetaMask team config.


.eslintignore
```diff
@@ -0,0 +1 @@
+!.eslintrc.js
```
Gudahtt: Curious - why is this ignored?

whymarrh: It's negated to allow ESLint to run against the config file itself, it doesn't by default. I want to make sure the config file itself adheres to the ruleset.

Gudahtt: Oh I see, neat

PR #16, opened 2020-01-29
Use bootstrap styles in 3-column layout

Gudahtt:
This is sort of a combination of both #14 and #15

### DIFF
--- a/metamask.css
+++ b/metamask.css
@@ -3,6 +3,10 @@ body {
   margin: 20px;
 }
 
+section {
+  margin: 20px 0 20px 0;
+}
+
 section > div.buttons {
   display: flex;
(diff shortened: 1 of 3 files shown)


metamask.css
```diff
@@ -0,0 +1,26 @@
...
+  margin: 20px 0 20px 0;
+}
+
+section > div.buttons {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+}
+
+.buttons > button.btn {
```
Gudahtt: This was the tangly part. I had to increase the specificy of this block by referencing a class and a parent class, to ensure it'd override the Bootstrap styles.

PR #29, opened 2020-04-14
fix for ethereum not found when metamask not installed

jennypollack:


contract.js
```diff
@@ -33,6 +33,8 @@ The `piggybankContract` is compiled from:
 const forwarderOrigin = 'http://localhost:9010'
 
 const isMetaMaskInstalled = () => {
+
+  const { ethereum } = window
```
rekmarks, line 37: nit: remove newline
```suggestion
  const { ethereum } = window
```

PR #30, opened 2020-04-16
Update account/initialization handling

rekmarks:
- Call `eth_accounts` on initialization, in case the initial `accountsChanged` is emitted before `DOMContentLoaded`, and therefore before the listener is set up
  - This is a general recommendation for dapps, btw (to be added to our docs)
- Change `ethereum.enable()` to `ethereum.send('eth_requestAccounts')`, and handle the returned accounts
- Minor refactoring

### DIFF
--- a/contract.js
+++ b/contract.js
@@ -36,6 +36,7 @@ const isMetaMaskInstalled = () => {
   const { ethereum } = window
   return Boolean(ethereum && ethereum.isMetaMask)
 }
+
 // Dapp Status Section
 const networkDiv = document.getElementById('network')
 const chainIdDiv = document.getElementById('chainId')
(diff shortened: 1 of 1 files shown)


contract.js
```diff
@@ -102,7 +104,8 @@ const initialize = () => {
 
   const onClickConnect = async () => {
     try {
-      await ethereum.enable()
+      accounts = await ethereum.send('eth_requestAccounts')
+      handleNewAccounts(accounts)
```
Gudahtt, line 108: Nit: this is assigning the result to `accounts` here, then assigning it again in `handleNewAccounts`

Maybe it should be a local variable instead? e.g.

```suggestion
      const newAccounts = await ethereum.send('eth_requestAccounts')
      handleNewAccounts(newAccounts)
```

PR #31, opened 2020-04-20
Use new EIP 1193 API

rekmarks:
Uses `ethereum.request` to make RPC requests instead of the now-deprecated `sendAsync`.


website/index.js
```diff
@@ -403,13 +403,14 @@ const initialize = () => {
...
 
-    ethereum.sendAsync({ method: 'eth_accounts' }, (err, response) => {
-      if (err) {
-        console.error('Error on init when getting accounts', err)
-      } else {
-        handleNewAccounts(response.result)
-      }
-    })
+    try {
+      const newAccounts = await ethereum.request({
```
rekmarks, line 407: This is why `initialize` is now `async`.

PR #33, opened 2020-04-28
Prepare to publish as @metamask/test-dapp

rekmarks:
- Rename files
  - Update deploy script
- Update `package.json`
  - name, files


Gudahtt:
The `www` directory seems a bit... unconventional? I'm also not sure why those files would be in it, and not `index.html` and `index.js`

PR #39, opened 2020-04-30
Fix html formatting; add prettier

rekmarks:
This PR has been tested on the extension with the `metamask-ui.spec` e2e test, locally and in CI. It works.

- Fix HTML formatting to make it more readable, and to ensure button `text` values match expectations and are selectable via XPath
  - See [this comment](https://github.com/MetaMask/test-dapp/pull/39#discussion_r418202602) for details
  - Add `prettier` to lint scripts in support of this goal


Gudahtt:
>ensure button text values match expectations and are selectable via XPATH

It's hard to tell what this change refers to, because most of the diff is due to `prettier`

website/index.html
```diff
@@ -1,102 +1,236 @@
...
-            <div class="card-body">
-              <h4 class="card-title">Send Tokens</h4>
-              <p class="info-text alert alert-success">Token: <span id="tokenAddress"></span></p>
-              <button class="btn btn-primary btn-lg btn-block mb-3" id="createToken" disabled>Create Token</button>
-              <button class="btn btn-primary btn-lg btn-block mb-3" id="transferTokens" disabled>Transfer Tokens</button>
-              <button class="btn btn-primary btn-lg btn-block mb-3" id="approveTokens" disabled>Approve Tokens</button>
-              <button class="btn btn-primary btn-lg btn-block mb-3" id="transferTokensWithoutGas" disabled>Transfer Tokens
-                Without Gas</button>
-              <button class="btn btn-primary btn-lg btn-block mb-3" id="approveTokensWithoutGas" disabled>Approve Tokens
-                Without Gas</button>
```
rekmarks, line 90: @Gudahtt while this appears as `Approve Tokens Without Gas`, in reality the value was something like `Approve Tokens                      Wihtout Gas`.

website/index.html
```diff
@@ -1,102 +1,236 @@
...
+                <h4 class="card-title">
+                  Basic Actions
+                </h4>
+
+                <button
+                  class="btn btn-primary btn-lg btn-block mb-3"
+                  id="connectButton"
+                  disabled
+                >
+                  Connect
```
Gudahtt, line 72: Why was `Connect` added here? This was removed because we'd determine during the page load what to display here (depending on whether they were connected and whether MetaMask was installed)

rekmarks: That still works, I just figured it was nice and potentially informative to have a default.

Gudahtt: Sure, I suppose so

website/index.html
rekmarks: Added this example to the README, too.

PR #41, opened 2020-05-08
Fix locally serving onboarding bundle

rekmarks:
Recent work accidentally broke the import of the onboarding bundle when serving the site locally. This PR fixes this by means of a `postinstall` script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "lint": "eslint . --ext js,json && prettier ./website -c",
     "lint:fix": "eslint . --fix --ext js,json && prettier ./website -c --write",
     "start": "static-server ./website --port 9011",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -7,6 +7,7 @@
     "lint": "eslint . --ext js,json && prettier ./website -c",
     "lint:fix": "eslint . --fix --ext js,json && prettier ./website -c --write",
     "start": "static-server ./website --port 9011",
+    "postinstall": "cp ./node_modules/@metamask/onboarding/dist/metamask-onboarding.bundle.js ./website",
```
Gudahtt, line 10: This seems broken

rekmarks: It works on my system 🤔

PR #46, opened 2020-05-12
Fix deploy script

rekmarks:
Because a release of this package wouldn't be complete if I didn't break the deploy script.


deploy.sh
```diff
@@ -50,9 +50,6 @@ function preprocess_and_publish {
     abort "git rev-parse returned an invalid shorthash from branch '${SOURCE_BRANCH}'"
   fi
 
-  # commit to destination branch with shorthash in message
```
rekmarks, line 53: Since we're no longer modifying any files (specifically, replacing the onboarding script `src`), there's nothing to commit, and this always fails.

PR #53, opened 2020-05-26
Update prepare script

rekmarks:
- Fix `[object Object]` error printout
- Move `prepare` commands into new file: `prepare.sh`
- Add `prepare` to `deploy` script

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -u
+set -e
+set -o pipefail
+
+mkdir -p ./website/scripts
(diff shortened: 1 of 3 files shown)


prepare.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -u
+set -e
+set -o pipefail
+
+mkdir -p ./website/scripts
+rm -rf ./website/scripts/*
```
rekmarks, line 8: - The folder may already exist, hence `mkdir -p`
- If the folder already exists, there may be stuff we don't want in it, and we should make sure to delete it

PR #76, opened 2020-08-07
Use ethers

rekmarks:
- Replace usage of `web3` with `ethers` everywhere
- Remove `web3`


Gudahtt:
@rekmarks This needs to be rebased, when you get the chance!

Gudahtt (approved):
LGTM!

PR #78, opened 2020-08-21
[WIP] Refactoring and Partitioning the 'Actions' into Modules

BboyAkers:
Partitioning the logic into their own separate modules will make it easier to extend/implement more code into the `index.js`


BboyAkers:
Will re-open when I can dedicate a substantial amount of time to refactoring.

PR #80, opened 2020-10-20
Use @metamask/eslint-config@4.0.0

whymarrh:
This change updates the shared ESLint config to the latest published version, v4.0.0, and fixes one new issue. The `.engines` property has been added to the `package.json` file to ensure that the Node.js-specific rules are targeting the correct version of the runtime.


Gudahtt (approved):
LGTM!

PR #83, opened 2020-10-28
Migrating Signature examples

NiranjanaBinoy:
Current changes are printing the results of the test to console.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,9 @@
+/* eslint-disable linebreak-style */
 import MetaMaskOnboarding from '@metamask/onboarding'
-import { encrypt } from 'eth-sig-util'
+// eslint-disable-next-line camelcase
+import { encrypt, recoverPersonalSignature, recoverTypedSignatureLegacy, recoverTypedSignature, recoverTypedSignature_v4 } from 'eth-sig-util'
 import { ethers } from 'ethers'
+import { toChecksumAddress } from 'ethereumjs-util'
 import { hstBytecode, hstAbi, piggybankBytecode, piggybankAbi } from './constants.json'
 
 // We must specify the network as 'any' for ethers to allow network changes
@@ -74,6 +77,15 @@ const encryptionKeyDisplay = document.getElementById('encryptionKeyDisplay')
 const ciphertextDisplay = document.getElementById('ciphertextDisplay')
 const cleartextDisplay = document.getElementById('cleartextDisplay')
 
+// Ethereum Signature Secrtion
+const ethSign = document.getElementById('ethSign')
+const ethSignResult = document.getElementById('ethSignResult')
+const personalSignTest = document.getElementById('personalSignTest')
+const personalRecoverTest = document.getElementById('personalRecoverTest')
+const signTypedDataTest = document.getElementById('signTypedDataTest')
+const signTypedDataV3Test = document.getElementById('signTypedDataV3Test')
+const signTypedDataV4Test = document.getElementById('signTypedDataV4Test')
+
 const initialize = async () => {
 
   let onboarding
@@ -102,6 +114,12 @@ const initialize = async () => {
     encryptMessageInput,
     encryptButton,
     decryptButton,
+    ethSign,
+    personalSignTest,
+    personalRecoverTest,
+    signTypedDataTest,
+    signTypedDataV3Test,
+    signTypedDataV4Test,
   ]
 
   const isMetaMaskConnected = () => accounts && accounts.length > 0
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -473,6 +496,256 @@ const initialize = async () => {
...
+      if (recoveredAddr === from) {
+        console.log(`SigUtil Successfully verified signer as ${recoveredAddr}`)
+      } else {
+        console.log(`SigUtil Failed to verify signer when comparing ${recoveredAddr} to ${from}`)
+        console.log(`Failed comparing ${recoveredAddr} to ${from}`)
+      }
+    } catch (err) {
+      console.error(err)
+    }
+  }
```
rekmarks, line 545: We should get rid of this functionality (the function, all related buttons, HTML etc.) and just use `personalRecoverTest` instead.

src/index.js
```diff
@@ -473,6 +496,256 @@ const initialize = async () => {
...
+
+  /**
+   * Personal ecRecover Test
+   */
+  personalRecoverTest.onclick = async () => {
```
rekmarks, line 550: This should be renamed to `personalSign` and replace the existing `personalSign` implementation in this file.

src/index.js
```diff
@@ -473,6 +496,256 @@ const initialize = async () => {
...
+    const exampleMessage = 'Sample message for personal ec Recover Test'
```
rekmarks, line 551: This should use the existing `personalSign` example message:
```suggestion
      const exampleMessage = 'Example `personal_sign` message'
```

src/index.js
```diff
@@ -74,6 +76,15 @@ const encryptionKeyDisplay = document.getElementById('encryptionKeyDisplay')
 const ciphertextDisplay = document.getElementById('ciphertextDisplay')
 const cleartextDisplay = document.getElementById('cleartextDisplay')
 
+// Ethereum Signature Secrtion
```
rekmarks, line 79: Typo:
```suggestion
// Ethereum Signature Section
```

src/index.js
```diff
@@ -473,6 +496,256 @@ const initialize = async () => {
...
+      }
+    } catch (err) {
+      console.log(err)
+    }
+  }
+
+  /**
+   * Sign Typed Data Version 3 Test
+   */
+  signTypedDataV3Test.onclick = async () => {
```
rekmarks, line 618: `signTypedDataV3Test` is redundant with the existing `signTypedData` button in this file. We should remove `signTypedDataV3Test` completely. And rename the existing `signTypedData` button to `signTypedDataV3`.

src/index.js
```diff
@@ -473,6 +496,256 @@ const initialize = async () => {
...
+        console.log(`Failed to verify signer when comparing ${ecRecoverAddr} to ${from}`)
+      }
+    } catch (err) {
+      console.error(err)
+    }
+  }
+
+  /**
+   * Sign Typed Data Test
+  */
```
rekmarks, line 577: Formatting issue:
```suggestion
  /**
   * Sign Typed Data Test
   */
```

src/index.js
```diff
@@ -473,6 +496,256 @@ const initialize = async () => {
...
+      ethSignResult.innerHTML = JSON.stringify(ethResult)
+    } catch (err) {
+      console.log(err)
+    }
+  }
+
+  /**
+     * Personal Sign
+     */
+
```
rekmarks, line 521: Formatting issue:
```suggestion
  /**
   * Personal Sign
   */
```

src/index.js
```diff
@@ -473,6 +496,256 @@ const initialize = async () => {
     }
   }
 
+  /**
+   * Ethereum Signature
```
rekmarks, line 500: Let's just use the RPC method names:
```suggestion
   * eth_sign
```

src/index.js
rekmarks: We can recover the signature twice in `personalRecoverTest`: once using `personal_ecRecover`, and once using `receoverPersonalSignature`.

rekmarks (changes requested):
Great work!

I left a number of inline comments requesting changes. Beyond those, we also want to reorganize the HTML somewhat. I'll explain what we should do in a follow-up.

Note that anything that's renamed in `index.js` should also be renamed in `index.html`, as appropriate.

rekmarks:
For reorganizing the HTML, first of all, we should move `Send Eth` to the `Contract` card, with `Send Eth` on top:

test_dapp_changes_1

rekmarks:
For the second HTML reorganization, after all other changes have been addressed, all signature methods should be grouped into two cards, as follows:

1. `eth_sign`, `personal_sign`
2. All `signTypedData` methods

Because we use the test dapp in our e2e tests, and it selects via HTML and CSS, please keep changes to existing HTML and CSS attributes/names/IDs/classes to a minimum!

rekmarks (changes requested):
Looking good! We're almost done here.

I made a number of suggested changes in specific places. Some of them should be applied everywhere, and I tried to indicate that in the comment.

In addition, all of the signature errors that are logged to console should be logged using `console.error` rather than `console.log`.

rekmarks (approved):
Great job!

PR #85, opened 2020-11-04
Skip lint and build upon install

Gudahtt:
Previously `yarn lint` and `yarn build` would get called upon install. I found this to be unnecessary and inconvenient. The `prepare` script that caused this has been replaced with a `prepublishOnly` script, so that it still runs before publish but not upon each install.

The old `prepare` script was also used during deployment, though worryingly we didn't handle the case where it fails. `yarn lint` and `yarn build` are now called separately during the deploy script, and we handle failures for either.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -53,7 +53,8 @@ function get_shorthash {
 
 function preprocess_and_publish {
 
-  yarn prepare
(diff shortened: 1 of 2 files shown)


rekmarks (approved):
Yeah, you're right!

PR #86, opened 2020-11-04
Fix onboarding

Gudahtt:
With the migration to `ethers`, we accidentally broke onboarding. We had assumed that `window.ethereum` was always present.

The initialization steps that use `window.ethereum` have been moved into the `initialization` method, and wrapped in a `try..catch` to ensure te remainder of the intialization proceeds even if MetaMask is not installed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,20 +5,9 @@ import { ethers } from 'ethers'
 import { toChecksumAddress } from 'ethereumjs-util'
 import { hstBytecode, hstAbi, piggybankBytecode, piggybankAbi } from './constants.json'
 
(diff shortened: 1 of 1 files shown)


Gudahtt:
In attempting to test this, I discovered there is a separate bug in the extension as well :facepalm: . I confirmed that with these changes, the event at least made its way to the extension background.

rekmarks (approved):

Gudahtt:
Here is the extension onboarding bug I referred to earlier, for reference: https://github.com/MetaMask/metamask-extension/issues/9792

PR #91, opened 2021-01-13
Fix `remote` option of deploy script

Gudahtt:
The `remote` option for the deploy script would not correctly set the remote used for deployment. It is used for everything but the most important step: the deploy step. For the `git push` that ultimately deploys the new build, the remote origin was hard-coded as the MetaMask test dapp repo.

A git command is now used to dynamically fetch the remote URL using the remote name given instead.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -59,12 +59,15 @@ function preprocess_and_publish {
   local shorthash
   shorthash=$(get_shorthash)
 
+  local remote_url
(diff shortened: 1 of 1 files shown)


Gudahtt:
I've just tested this with my fork - the deploy script does seem to work correctly now with this change.

rekmarks (approved):

PR #93, opened 2021-03-10
Fix addEthereumChain button disable logic

rekmarks:
The button should be disabled if MetaMask is not installed, just like every other button.


Gudahtt (dismissed):
LGTM!

Gudahtt (approved):

PR #94, opened 2021-03-12
Replace 'master' references with 'main'

rekmarks:
Replaces references to `master` with `main` following default branch renaming.


Gudahtt (approved):
LGTM!

PR #97, opened 2021-03-12
Fix deploy script by parameterizing init branch

rekmarks:
Our deploy script was relying on the author and the repo having the same default git branch name. This parameterizes the `git init` default branch in the deploy script to ensure that it works regardless of the repository's default branch name and the deployer's default branch name.


ryanml (approved):
++

rekmarks:
CircleCI is having trouble, but this PR is good, so I'm going to admin-merge it.

PR #118, opened 2021-06-25
Fix repository standardization issues

rekmarks:
- Add `@lavamoat/allow-scripts`, update readme
- Update ESLint dependencies, config, and scripts


brad-decker (approved):

.eslintrc.js
```diff
@@ -1,28 +1,27 @@
...
+  extends: ['@metamask/eslint-config', '@metamask/eslint-config-nodejs'],
+
+  overrides: [
+    {
+      files: ['src/*.js'],
+      parserOptions: {
+        sourceType: 'module',
+      },
+      rules: {
+        camelcase: ['error'],
```
Gudahtt, line 21: :eyes: 

Why is this listed? The default camelcase rule should work fine here.

.eslintrc.js
```diff
@@ -1,28 +1,27 @@
...
+      },
```
rekmarks, line 22:
```suggestion
```

.eslintrc.js
rekmarks: Doh, I had overrides that I removed, and forgot to remove the rule itself. Fixed in 8683953

Gudahtt (approved):
LGTM!

PR #119, opened 2021-07-26
Add EIP1559 send that does not suggest any fees

danjm:


darkwing (approved):

src/index.js
```diff
@@ -335,7 +338,7 @@ const initialize = async () => {
           {
             from: accounts[0],
             to: '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
-            value: '0x29a2241af62c0000',
+            value: '0x0',
```
Gudahtt, line 341: Was this change related? :thinking: Not that I'm opposed - it just seems out of place.

PR #125, opened 2021-09-04
Improve encrypt/decrypt flow

### LINKED ISSUE #124: Make the "encrypt/decrypt" section usable as a minimal messenger
Current decryption demo only allows decrypting the message you encrypted to yourself. This doesn't really show off the feature.

The sections should be like this:

```
[get encryption key]
encryption key output
-------
type message
recipient encryption key
[encrypt message]
encrypted output (full, copyable)
--------
type encrypted input
[decrypt message]
decrypted output
```

danfinlay:
Made encrypt/decrypt a "fully functional" flow, so the test dapp can
actually be used as an encrypted messenger.

Fixes #124

Looks like this:
image

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -258,12 +258,17 @@
                 <button
                   class="btn btn-primary btn-lg btn-block mb-3"
                   id="getEncryptionKeyButton"
-                  disabled
                 >
-                  Get Encryption Key
+                  Get Public Encryption Key
                 </button>
 
+                <p class="info-text alert alert-secondary">
+                  Public Encryption key: <span id="encryptionKeyDisplay"></span>
+                </p>
                 <hr />
+                <h5 class="card-title">
+                  Encrypt
+                </h5>
 
                 <div id="encrypt-message-form">
                   <input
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -258,12 +258,17 @@ <h4 class="card-title">
                 <button
                   class="btn btn-primary btn-lg btn-block mb-3"
                   id="getEncryptionKeyButton"
-                  disabled
```
Gudahtt, line 261: Why was this removed? The intention was to have all of the buttons that rely upon an account being connected to start as disabled until we know whether something is connected or not.

src/index.html
```diff
@@ -272,33 +277,45 @@ <h4 class="card-title">
...
+                    class="form-control"
+                    type="text"
+                    placeholder="Recipient's Public Encryption Key"
+                    id="publicKeyInput"
+                  />
 
                   <button
                     class="btn btn-primary btn-lg btn-block mb-3"
                     id="encryptButton"
-                    disabled
```
Gudahtt, line 279: Similarly, this one was meant to be disabled until we have a public encryption key to encrypt with.

src/index.js
```diff
@@ -6,6 +6,7 @@ import {
   recoverTypedSignatureLegacy,
   recoverTypedSignature,
   recoverTypedSignature_v4 as recoverTypedSignatureV4,
+  decrypt,
```
Gudahtt, line 9: This line is causing the linter to fail

src/index.html
Gudahtt: The same goes for the "decrypt" button.

src/index.js
```diff
@@ -511,32 +518,17 @@ const initialize = async () => {
       }
     };
 
-    encryptMessageInput.onkeyup = () => {
-      if (
-        !getEncryptionKeyButton.disabled &&
-        encryptMessageInput.value.length > 0
-      ) {
-        if (encryptButton.disabled) {
-          encryptButton.disabled = false;
```
Gudahtt, line 520: Was this deleted to allow encrypting an empty message? I'm not sure why that was disallowed to being with :thinking: Does that work?

Gudahtt (changes requested):
This design looks good! This will almost certainly break the e2e tests, but we can fix those easily enough. We should publish this as a breaking change to indicate that.

I think the button disabled state logic needs some cleanup. For the two new inputs (public encryption key and encrypted message), we could add listeners to enable the associated button when they are non-empty.

PR #126, opened 2021-09-30
Add type-0 to legacy transaction

jpuri:
Adding type `type: '0x0'` to legacy transaction.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -319,11 +319,18 @@ const initialize = async () => {
      */
 
     sendButton.onclick = async () => {
-      const result = await ethersProvider.getSigner().sendTransaction({
-        to: '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
-        value: '0x29a2241af62c0000',
-        gasLimit: 21000,
-        gasPrice: 20000000000,
+      const result = await ethereum.request({
+        method: 'eth_sendTransaction',
+        params: [
+          {
+            from: accounts[0],
+            to: '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
+            value: '3782DACE9D90000',
+            gasLimit: '0x5028',
+            gasPrice: '0x2540be400',
+            type: '0x0',
+          },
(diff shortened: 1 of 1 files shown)


Gudahtt (approved):
LGTM!

As a later enhancement, it would be neat if we could control the type used in the transaction, so that we can test what we'd see from a legacy dapp, or one that sends invalid params.

src/index.js
```diff
@@ -319,11 +319,18 @@ const initialize = async () => {
...
-        value: '0x29a2241af62c0000',
-        gasLimit: 21000,
-        gasPrice: 20000000000,
+      const result = await ethereum.request({
+        method: 'eth_sendTransaction',
+        params: [
+          {
+            from: accounts[0],
+            to: '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
+            value: '3782DACE9D90000',
```
Gudahtt, line 328: Oh, huh, why the change in `value`?

jpuri: previous values of 3 eth was emptying my account 😄

rekmarks: This is a most welcome change! Be advised that it may break our e2e tests, but yeah, 3 ETH is too much, and I always end up editing the amount in practice.

One thing, we should `0x`-prefix it for consistency, and if we're changing it, we might as well go even further to 0.1 ETH, as opposed to the current 0.25:

```suggestion
            value: '0x16345785d8a0000',
```

jpuri: @PeterYinusa : please note it may need change in e2e tests for legacy transactions.

Gudahtt: That's what I was going to say as well - this isn't a valid parameter without the "0x" prefix.

See here: https://eth.wiki/json-rpc/API#hex-value-encoding
This value is a QUANTITY, the rules for which are described there.

Gudahtt (approved):
LGTM!

darkwing (approved):

PR #129, opened 2021-10-14
Do not send to hacked address, and send 0 eth.

danfinlay:
We should make this send to the user's own address, not to [a known hacked address that gets swept](https://etherscan.io/address/0x2f318C334780961FB129D2a6c30D0763d9a5C970). It's so scary that we were suggesting 3 eth transactions to a hacked address in our test dapp!

As a short term patch, I am directing this to the team account.


Gudahtt (approved):
LGTM!

PR #133, opened 2021-11-04
Fixes signTypedData_v3 message

alex-miller-0:
The previous message was not compilant with EIP712 and was rejected by
the GridPlus Lattice. The `message` key names must map to the `Mail`
object definitions.


Gudahtt (approved):
LGTM!

PR #136, opened 2021-11-19
add collectibles flow

PeterYinusa:
- add collectibles section
- deploy contract
- mint collectible(s)

Opensea: https://testnets.opensea.io/collection/testdappcollectibles

opensea

Etherscan: https://rinkeby.etherscan.io/token/0x69cd64c642329943e5806d4f99c39c6c8890437d
etherscan

Collectibles display:

Collectibles Display


src/index.js
```diff
@@ -329,6 +349,46 @@ const initialize = async () => {
       console.log(contract);
     };
 
+    /**
+     * ERC721 Token
+     */
+
+    deployCollectiblesButton.onclick = async () => {
```
mcmire, line 356: It's better practice to use `addEventListener` rather than set `onclick` directly, in case we want to add other listeners to this in the future. So I would recommend:

``` suggestion
deployCollectiblesButton.addEventListener('click', () => {
```

src/index.js
```diff
@@ -329,6 +349,46 @@ const initialize = async () => {
...
+      }
+
+      console.log(
+        `Contract mined! address: ${contract.address} transactionHash: ${contract.transactionHash}`,
+      );
+      collectiblesStatus.innerHTML = 'Deployed';
+      mintButton.disabled = false;
+      mintAmountInput.disabled = false;
+
+      mintButton.onclick = async () => {
```
mcmire, line 379: Same thing as above:

``` suggestion
mintButton.addEventListener('click', () => {
```

mcmire (approved):
Have a couple of comments but this seems reasonable to me.

src/index.js
PeterYinusa: will deal with that in a separate PR

src/index.js
PeterYinusa: and this too

darkwing (approved):

Issue #18, opened 2020-02-27
Implement batch transactions


rekmarks:
We should add batch transaction functionality to the test dapp, for testing.

We should follow this:
https://codesandbox.io/s/metamask-batchrequest-bug-demo-emnph

Zipped:
[metamask-batchrequest-bug-demo.zip](https://github.com/MetaMask/test-dapp/files/4263917/metamask-batchrequest-bug-demo.zip)

david0xd:
@danjm During the refinement we made a conclusion that the best approach to achieve this is to create a new button on the test-dapp which will call wrapped function that will perform few transactions in a batch which will target our deployed smart contract.

AC:
* Add new button "Deposit in batch" under the "Contract" block
* Make sure that the new button is available only when the contract is deployed (like the other buttons in that section)
* Link new button to the new function that will perform 4 batched transactions to deposit ETH to the deployed smart contract

Couple of arguments for the proposal is:
* To make it adapted to the current test dapp web page and keep it simple for use
* We would probably like to use our smart contract that we deployed inside the test dapp instead of configuring an external

danjm:
This proposal sounds good to me. Are there any other blockers to this?

EHaracic:
Blocked by PR: https://github.com/MetaMask/test-dapp/issues/137

david0xd:
Because of limitation of ethers library we've decided to integrate web3 library into the test-dapp. Web3 library with its functionality will make batch transactions possible.
This can be continued when https://github.com/MetaMask/test-dapp/pull/146 this one gets merged.

PR #146, opened 2021-12-23
Added radio button toggle to use Web3 provider to send transaction

VSaric:
fixes: #137 

**Description:** Added a radio button toggle on the top of the page. The options in the toggle is `ethers` and `web3`.  Default toggle is `ethers`. If `ethers` is selected, the create token and deploy contract buttons are behave as they do today. If `web3` is selected, create token and deploy contract use `web3` instead of the current `factories` created by the `ethers` provider. The transactions created by the buttons are the same. 

**Screenshot of radio button toggle:**

Screenshot from 2021-12-24 10-46-31

**Switching ethers/web3 sreencast:** 

https://user-images.githubusercontent.com/92527393/147463129-c8e42464-1aa5-41b8-92be-da08ea57986b.mp4

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -43,7 +43,7 @@ input[type='radio'] {
   margin: 0 5px 0 5px;
 }
 
-#libraryLabel {
+.libraryLabel {
   font-size: large;
 }
 
--- a/src/index.css
+++ b/src/index.css
@@ -36,6 +36,17 @@ header {
   width: 100%;
 }
 
+/* Radio buttons switch */
+
+input[type='radio'] { 
+  transform: scale(1.5); 
+  margin: 0 5px 0 5px;
+}
+
+#libraryLabel {
+  font-size: large;
+}
+
 /* Encrypt/Decrypt */
 
 #encryptMessageInput {
--- a/src/index.html
+++ b/src/index.html
@@ -30,6 +30,16 @@
         </div>
       </header>
       <section>
+        <form name="myForm" class="form">
+          <div class="row justify-content-center">
+            <div class="justify-content-center">
+              <input type="radio" id="radio-ethers" name="libraryUsed" value="ethers" />
+              <label id="libraryLabel">Ethers</label>
+              <input type="radio" id="radio-web3" name="libraryUsed" value="web3" />
+              <label id="libraryLabel">Web3</label>
+            </div>
+          </div>
+        </form>
         <h3 class="card-title">
           Status
         </h3>
--- a/src/index.html
+++ b/src/index.html
@@ -30,13 +30,13 @@
         </div>
       </header>
       <section>
-        <form name="myForm" class="form">
+        <form name="librarySwitchForm" class="form">
           <div class="row justify-content-center">
             <div class="justify-content-center">
               <input type="radio" id="radio-ethers" name="libraryUsed" value="ethers" />
-              <label id="libraryLabel">Ethers</label>
+              <label class="libraryLabel">Ethers</label>
               <input type="radio" id="radio-web3" name="libraryUsed" value="web3" />
-              <label id="libraryLabel">Web3</label>
+              <label class="libraryLabel">Web3</label>
             </div>
           </div>
         </form>
--- a/src/index.js
+++ b/src/index.js
@@ -192,22 +192,14 @@ function initializeLibrarySwitches() {
 function handleLibraryChange(event) {
   libraryInUse = event.target.value;
   if (libraryInUse === 'web3') {
-    // Deployed piggy bank smart contract with web3 library
     piggyBankContract = web3PiggyBankContract;
-    // Deployed hst smart contract with web3 library
     hstContract = web3HstContract;
-    // Deployed collectibles smart contract with web3 library
     collectiblesContract = web3CollectiblesContract;
-    // Deployed failing smart contract with web3 library
     failingContract = web3FailingContract;
   } else {
-    // Deployed piggy bank smart contract with ethers library
     piggyBankContract = ethersPiggyBankContract;
-    // Deployed hst smart contract with ethers library
     hstContract = ethersHstContract;
-    // Deployed collectibles smart contract with ethers library
     collectiblesContract = ethersCollectiblesContract;
-    // Deployed failing smart contract with ethers library
     failingContract = ethersFailingContract;
   }
 
@@ -217,7 +209,7 @@ function handleLibraryChange(event) {
     depositButton.disabled = false;
     withdrawButton.disabled = false;
   } else {
-    contractStatus.innerHTML = 'Not clicked';
+    contractStatus.innerHTML = 'Not deployed';
     depositButton.disabled = true;
     withdrawButton.disabled = true;
   }
@@ -258,11 +250,9 @@ function handleLibraryChange(event) {
     failingContractStatus.innerHTML = 'Deployed';
     sendFailingButton.disabled = false;
   } else {
-    failingContractStatus.innerHTML = 'Not clicked';
+    failingContractStatus.innerHTML = 'Not deployed';
     sendFailingButton.disabled = true;
   }
-
-  console.log(libraryInUse);
 }
 
 const initialize = async () => {
@@ -499,8 +489,12 @@ const initialize = async () => {
             })
             .then((receipt) => {
               console.log(receipt);
+              contractStatus.innerHTML = 'Deposit completed';
+            })
+            .catch((error) => {
+              console.log(error);
+              contractStatus.innerHTML = 'Contract call error';
             });
-          contractStatus.innerHTML = 'Deposit completed';
         };
       }
 
@@ -524,12 +518,14 @@ const initialize = async () => {
             })
             .then((receipt) => {
               console.log(receipt);
+              contractStatus.innerHTML = 'Withdrawn';
+            })
+            .catch((error) => {
+              console.log(error);
+              contractStatus.innerHTML = 'Contract call error';
             });
-          contractStatus.innerHTML = 'Withdrawn';
         };
       }
-
-      console.log(piggyBankContract);
     };
 
     deployFailingButton.disabled = false;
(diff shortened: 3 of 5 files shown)


src/index.html
```diff
@@ -30,6 +30,16 @@ <h1 id="logo-text" class="text-center">
         </div>
       </header>
       <section>
+        <form name="myForm" class="form">
```
david0xd, line 33: This name can be something like "librarySwitchForm".

src/index.css
```diff
@@ -36,6 +36,17 @@ header {
...
 }
 
+/* Radio buttons switch */
+
+input[type='radio'] { 
+  transform: scale(1.5); 
+  margin: 0 5px 0 5px;
+}
+
+#libraryLabel {
```
david0xd, line 46: This should be a class instead of an ID selector -> `.libraryLabel`.

src/index.html
```diff
@@ -30,6 +30,16 @@ <h1 id="logo-text" class="text-center">
...
+          <div class="row justify-content-center">
+            <div class="justify-content-center">
+              <input type="radio" id="radio-ethers" name="libraryUsed" value="ethers" />
+              <label id="libraryLabel">Ethers</label>
```
david0xd, line 37: Replace `id` with `class` here for both labels.

src/index.js
```diff
@@ -142,7 +147,149 @@ const submitFormButton = document.getElementById('submitForm');
...
+    radioEthers.checked = true;
+  } else {
+    radioWeb3.checked = true;
+  }
+}
+
+function handleLibraryChange(event) {
+  libraryInUse = event.target.value;
+  if (libraryInUse === 'web3') {
+    // Deployed piggy bank smart contract with web3 library
```
david0xd, line 195: We can remove these comments for this particular section because code already looks self-explainable.

src/index.js
```diff
@@ -142,7 +147,149 @@ const submitFormButton = document.getElementById('submitForm');
...
+    piggyBankContract = web3PiggyBankContract;
+    // Deployed hst smart contract with web3 library
+    hstContract = web3HstContract;
+    // Deployed collectibles smart contract with web3 library
+    collectiblesContract = web3CollectiblesContract;
+    // Deployed failing smart contract with web3 library
+    failingContract = web3FailingContract;
+  } else {
+    // Deployed piggy bank smart contract with ethers library
```
david0xd, line 204: We can remove these comments for this particular section because code already looks self-explainable.

src/index.js
```diff
@@ -142,7 +147,149 @@ const submitFormButton = document.getElementById('submitForm');
...
+    failingContract = ethersFailingContract;
+  }
+
+  // Changing buttons and label state when piggy bank contract is deployed
+  if (piggyBankContract) {
+    contractStatus.innerHTML = 'Deployed';
+    depositButton.disabled = false;
+    withdrawButton.disabled = false;
+  } else {
+    contractStatus.innerHTML = 'Not clicked';
```
david0xd, line 220: This should probably say something like "Not deployed" instead of "Not clicked".

src/index.js
```diff
@@ -142,7 +147,149 @@ const submitFormButton = document.getElementById('submitForm');
...
+    transferTokensWithoutGas.disabled = true;
+    approveTokensWithoutGas.disabled = true;
+  }
+
+  // Changing buttons and label state when failing contract is deployed
+  if (failingContract) {
+    failingContractStatus.innerHTML = 'Deployed';
+    sendFailingButton.disabled = false;
+  } else {
+    failingContractStatus.innerHTML = 'Not clicked';
```
david0xd, line 261: This should probably say something like "Not deployed" instead of "Not clicked".

src/index.js
```diff
@@ -142,7 +147,149 @@ const submitFormButton = document.getElementById('submitForm');
...
+    sendFailingButton.disabled = true;
+  }
+
+  console.log(libraryInUse);
```
david0xd, line 265: This console log can be removed.

src/index.js
```diff
@@ -320,70 +467,74 @@ const initialize = async () => {
...
+          await piggyBankContract.methods
+            .deposit()
+            .send({
+              from: accounts[0],
+              value: '0x3782dace9d900000',
+            })
+            .then((receipt) => {
+              console.log(receipt);
+            });
+          contractStatus.innerHTML = 'Deposit completed';
```
david0xd, line 501: This should be moved in then() callback function instead of being here. 
Also, it would be nice to have catch on error and handle that as well (log to console and display status text with error message).

src/index.js
```diff
@@ -320,70 +467,74 @@ const initialize = async () => {
...
+        withdrawButton.onclick = async () => {
+          await piggyBankContract.methods
+            .withdraw('0xde0b6b3a7640000')
+            .send({
+              from: accounts[0],
+            })
+            .then((receipt) => {
+              console.log(receipt);
+            });
+          contractStatus.innerHTML = 'Withdrawn';
```
david0xd, line 524: This can also be moved to then() and error catch is welcome.

src/index.js
```diff
@@ -320,70 +467,74 @@ const initialize = async () => {
...
+        };
+      }
 
-      console.log(contract);
+      console.log(piggyBankContract);
```
david0xd, line 528: This can be removed.

VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.js
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.js
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.js
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.js
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.js
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.js
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.js
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.html
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.css
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

src/index.html
VSaric: Fixed in: 91555c04c8088ee472c87cc25843642994ec92b4

mirjanaKukic:
Verified by QA.

david0xd (approved):

VSaric:
Hello @danjm, we implemented the `web3` switching feature into the test-dapp and you can see it here in this PR. Just to remember you, if you have time, can you please review this changes, because when this PR be merged and done, we can unblocked ticket #18 and move it from `Pending Consensys` to `Ready for development` column.

Thank you in advance! :)

EHaracic:
fixes: #137

mcmire (dismissed):
I tested these changes, and they seem to work.

There's one thing about this PR that we need to think about. This will add some tech debt that I think we will need to resolve in order to be able to make it easy to maintain the test dapp in the future. First, after this PR, the size of `src/index.js` will be 1668 lines. It's not the most pleasant experience to maintain a file like this, and I think it'd be good if we split this up into separate files. Second, since we now have two ways of doing the same thing, I think we could benefit from using the strategy pattern to represent the logic between Web3 and ethers. That is, we could figure out all of the common actions between the two strategies and then create a Web3 object and an ethers object that implements the differences among all of the actions we've defined, then switch out the objects when the radio button is clicked.

I don't know if this change is time-sensitive, so if that's the case, I'm fine if we merge this now and refactor this later, but we're getting to a point now where the more we change, the more debt we create and the longer it will take to refactor. So it might be something we want to think about sooner rather than later.

david0xd:
@mcmire I agree with you, just the purpose of this was to unblock the https://github.com/MetaMask/test-dapp/issues/18 at the moment.
Even before web3 integration was started, that index.js requires some heavy refactoring.
So, we should create a new task to do some good refactoring, probably to split these blocks of functionalities into the separate files, do the switching and call-delegation at the higher level from the index.js and propagate forward.

@danjm can you consider creating new ticket for refactoring of the test-dapp (in complete)? I'm also wondering if this is a part of the TypeScript refactoring and should this also be TypeScript so we do all at once?

I would just suggest that we finish with this with the merge and implementation of the [batch transactions](https://github.com/MetaMask/test-dapp/issues/18) to unblock the requirements and then do the refactor.

david0xd (dismissed):

EHaracic:
@danjm can we merge this issue?

mcmire (approved):

david0xd (approved):

danjm:
I checked that this is compatible with extension e2e tests

danjm:
I think we are close to being able to merge this. Before we do I'd like to get a look at this from @PeterYinusa: Peter, does this meet the needs we have for avoiding web3 compatibility bugs, from a QA perspective?

PeterYinusa:
> I think we are close to being able to merge this. Before we do I'd like to get a look at this from @PeterYinusa: Peter, does this meet the needs we have for avoiding web3 compatibility bugs, from a QA perspective?

Yes, this looks good

mcmire (approved):
LGTM.

mcmire:
What was the final status on this PR, did we decide we didn't need it any more?

danjm:
The changes from this PR are going to be incorporated into future changes to the test dapp made by the dapp api / devrel team

PR #163, opened 2022-04-06
Fix event used for accessing provider

jpuri:
Using load event ensures that `window.ethereum` is always available in global scope when we access provider.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1271,7 +1271,7 @@ const initialize = async () => {
   }
 };
 
-window.addEventListener('DOMContentLoaded', initialize);
+window.addEventListener('load', initialize);
 
 // utils
 


Gudahtt:
The `load` event is not enough to guarantee that the provider is present on mobile. We have a library for blocking on the provider being injected (`@metamask/detect-provider`). Maybe that would be better here instead?

The reason the initialization was triggered by `DOMContentLoaded` is that the initialization _uses_ the DOM. It assumes all of the elements are on the page. Nothing happens on `load` that is directly tied to initialization.

Gudahtt:
Oh I see, but we do use `window.ethereum` right away during initialization. Hmm.

OK this seems like an improvement that we could further improve later by using `@metamask/detect-provider`, for better mobile compatibility.

Gudahtt (approved):
LGTM!

adonesky1 (approved):
LGTM!

PR #164, opened 2022-04-07
Add Sign-in with Ethereum Test Cases

skgbafa:
[(EIP-4361) Sign-in with Ethereum or SIWE](https://eips.ethereum.org/EIPS/eip-4361) is a standard message format that wallets can parse. In order to facilitate easy testing of some various test cases of this format, 4 cases are presented for signing with a wallet:

1. A normal SIWE message
2. A SIWE message with optional parameters included
3. A SIWE message with a different domain than the one sending the message (simulating a malicious use of SIWE)
4. A SIWE message with a different account than the one signing the message (simulating a error on the user's part)
5. A malformed SIWE message that doesn't meet the spec (missing a required field)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,25 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Added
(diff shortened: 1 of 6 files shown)


adonesky1:
Question about these: for the malicious/malformed cases should we be doing some validation in app?

BelfordZ:
@adonesky1 I think he has some of that going here: https://github.com/MetaMask/metamask-extension/pull/16616

BelfordZ (approved):
looks good. 

If/when extension adds better support for SIWE spec, the error cases will change for the better.

Issue #160, opened 2022-03-30
Deployed mode


PeterYinusa:
Load the Test Dapp in `Deployed` mode in order to interact with a deployed instance of each contract

Current mode:
- Deploy a Contract using the ContractFactory https://docs.ethers.io/v5/api/contract/contract-factory/

Deployed mode:
- Create an instance of a Contract https://docs.ethers.io/v5/api/contract/contract/

AC:
- [ ] User can toggle between the current mode and deployed mode
- [ ] Deployed mode uses a known instance of each contract

PR #180, opened 2022-06-23
Contract already deployed

PeterYinusa:
**Explanation:**

Fixes: https://github.com/MetaMask/test-dapp/issues/160

Updates the test dapp to consume a query string parameter that describes a deployed contracts address, such as `http://localhost:9011/?contract=0x7F775B6392f46Cf717aBB204557f64F68eB388c1`

This change will allow us to;
- avoid repeatedly deploying a contract when testing (manual testing)
- deploy a contract in the test setup instead of via the browser in the test steps (e2e tests)

**More information:**
- The contract address in the query string param will be applied to all contracts
- You can override the contract address for a specific contract by deploying it

**Before:**

After refreshing the page, there's no way to interact with the deployed contract

https://user-images.githubusercontent.com/53189696/175405985-4c510f99-0f6a-4acf-91b5-700ca6c024f2.mov

**After:**

Once you add the query string param, you can interact with a deployed contract

https://user-images.githubusercontent.com/53189696/175406016-bc77ff0d-e417-42b9-99e4-9e5e41e2e63a.mov

**Manual testing steps:**
- Without query string param: Test deploying and interacting with all contracts
- With query string param: Test interacting with all contracts

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -25,10 +25,16 @@ let hstFactory;
 let piggybankFactory;
 let collectiblesFactory;
 let failingContractFactory;
+let hstContract;
+let piggybankContract;
+let collectiblesContract;
+let failingContract;
 
 const currentUrl = new URL(window.location.href);
 const forwarderOrigin =
   currentUrl.hostname === 'localhost' ? 'http://localhost:9010' : undefined;
+const urlSearchParams = new URLSearchParams(window.location.search);
+const deployedContractAddress = urlSearchParams.get('contract');
 
 const { isMetaMaskInstalled } = MetaMaskOnboarding;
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -316,145 +367,142 @@ const initialize = async () => {
...
+    };
 
-      console.log(contract);
+    withdrawButton.onclick = async () => {
+      const result = await piggybankContract.withdraw('0xde0b6b3a7640000', {
+        from: accounts[0],
+      });
+      console.log(result);
+      const receipt = await result.wait();
+      console.log(receipt);
```
jpuri, line 414: Are these logs intentional ?

PR #188, opened 2022-08-15
Add release workflows

rickycodes:
This is the bare minimum setup for automating npm releases in this repository.

I've tested this in my fork here:
https://github.com/rickycodes/test-dapp/actions/runs/2864216714


Gudahtt:
It looks like the old deploy script is still present. Is that still needed?

legobeat (approved):
LGTM; @rickycodes anything outstanding you see as blocker for this currently?

Gudahtt:
The deploy step is still missing from this workflow. Which isn't a _complete_ blocker (some automation is better than none), but it should at least be called out in the instructions if it's to remain a manual step

PR #199, opened 2022-12-13
update instances of collectibles to nfts

NidhiKJha:
This PR is to update `collectibles `to `nfts`.

Reference: https://github.com/MetaMask/metamask-extension/pull/16889


src/constants.json
```diff
@@ -2,288 +2,352 @@
...01908152602001600020819055506110a9816000808573ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1681526020019081526020016000205461115590919063ffffffff16565b6000808473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020819055508173ffffffffffffffffffffffffffffffffffffffff168373ffffffffffffffffffffffffffffffffffffffff167fddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef836040518082815260200191505060405180910390a3505050565b600080828401905083811015151561116c57600080fd5b809150509291505056fea165627a7a723058205fcdfea06f4d97b442bc9f444b1e92524bc66398eb4f37ed5a99f2093a8842640029000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000c0000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000186a00000000000000000000000000000000000000000000000000000000000000003545354000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000035453540000000000000000000000000000000000000000000000000000000000",
```
mcmire, line 2: Can you explain the reason for this change? Was it just to simplify the contract being used here to produce NFTs? Or is there no real change and the keys here are just getting resorted?

src/index.html
```diff
@@ -312,12 +312,12 @@ <h4 class="card-title">
             <div class="card full-width">
               <div class="card-body">
                 <h4 class="card-title">
-                  Collectibles
+                  NFTS
```
mcmire, line 315: Nit: Maybe this should be `NFTs`

src/index.js
```diff
@@ -376,7 +374,7 @@ const initialize = async () => {
       multisigContractStatus.innerHTML = 'Deployed';
       sendMultisigButton.disabled = false;
       // ERC721 Token - Collectibles contract
```
Gudahtt, line 376: Nit: One leftover reference to "collectibles" here
```suggestion
      // ERC721 Token - NFT contract
```

src/constants.json
Gudahtt: If these sorting changes are unintentional, perhaps this file could be updated to only change the collectibles entries? That would make this easier to review.

NidhiKJha: My bad, this was not supposed to be updated. Only the collectiblesByteCode and collectiblesAbi have to be changed with nftsByteCode and nftsAbi.

contracts/TestDappNFTs.sol
```diff
@@ -1,25 +1,21 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.4.22 <0.9.0;
 
-import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
```
Gudahtt, line 4: Hmm, I tried compiling this in Remix but got an error about `Strings` already being declared. I don't quite understand why, but, it went away with I removed the `@4.8.1` or added it to the other two imports. Perhaps they should all be using the same version of `@openzeppelin/contracts`.

PR #202, opened 2022-12-16
Add interactions involving filters & subscriptions

mcmire:
We have a PR in the extension repo which touches code around RPC middleware for filter- and subscription-based RPC methods, and we want to be able to test these methods manually to ensure they are still working as designed. In order to accomplish this, I've added new cards which allow the user to:

* Create and remove a filter.
  * Currently there is support for creating a generic log filter via `eth_newFilter` as well as a block filter via `eth_newBlockFilter`. Once a filter is created, `eth_getFilterChanges` is then polled every 2 seconds. Filters are removed via `eth_uninstallFilter`.
  * I tried to add a button which created a pending transaction filter via `eth_newPendingTransactionFilter`, but it appears that `eth-json-rpc-filters` has a [bug](https://github.com/MetaMask/eth-json-rpc-filters/issues/81) which prohibits this RPC method from working fully.
* Start and stop a subscription.
  * As with filters, currently there is support for subscribing to new blocks via the `newHeads` parameter to `eth_subscribe` as well as new logs via the `logs` parameter. Subscriptions are stopped via `eth_unsubscribe`.
  * I also tried to add a button for subscribing to pending transactions, but [this doesn't seem to be supported outright by `eth-json-rpc-filters`](https://github.com/MetaMask/eth-json-rpc-filters/blob/5cbea3037b0655aa2c188d85b8ffe559a263dc0d/subscriptionManager.js#L50).

## Screencaps

https://user-images.githubusercontent.com/7371/207994523-eb96eae0-fa53-41fb-aa57-530e081696b4.mov

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,6 +20,200 @@ import {
   failingContractBytecode,
 } from './constants.json';
 
+function findElementById(id) {
+  const element = document.getElementById(id);
+
+  if (element === null) {
+    throw new Error(`Couldn't find element by id '${id}'`);
+  } else {
+    return element;
+  }
+}
+
+function setupFilters(filterType, filterChangesBoxPlaceholder) {
+  if (!['log', 'block', 'pendingTransaction'].includes(filterType)) {
+    throw new Error(
+      "filterType must be either 'log', 'block', or 'pendingTransaction'",
+    );
+  }
+
+  const capitalizedFilterType =
+    filterType[0].toUpperCase() + filterType.slice(1);
+  let filterId;
+  let pollingTimer;
+  const createFilterButton = findElementById(
+    `create${capitalizedFilterType}FilterButton`,
+  );
+  const removeFilterButton = findElementById(
+    `remove${capitalizedFilterType}FilterButton`,
+  );
+  const filterChangesBox = findElementById(`${filterType}FilterChangesBox`);
+
+  return {
+    createFilterButton,
+    removeFilterButton,
+    activateFilterButtons,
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -20,6 +20,200 @@ import {
...
+  } else {
+    return element;
+  }
+}
+
+function setupFilters(filterType, filterChangesBoxPlaceholder) {
+  if (!['log', 'block', 'pendingTransaction'].includes(filterType)) {
+    throw new Error(
+      "filterType must be either 'log', 'block', or 'pendingTransaction'",
+    );
```
legobeat, line 37: Mostly to get rid of double-quotes ;)
```suggestion
  const validFilterTypes = ['log', 'block', 'pendingTransaction'];
  if (!validFilterTypes.includes(filterType)) {
    throw new Error(
      `filterType must be one of: ${validFilterTypes.join(',')}.`,
    );
```

src/index.js
```diff
@@ -20,6 +20,200 @@ import {
...
+      stopSubscription().catch(console.error);
+    });
+  }
+
+  async function startSubscription() {
+    window.ethereum.addListener('message', onMessageReceived);
+
+    newSubscriptionId = await ethereum.request({
+      method: 'eth_subscribe',
+      params: [subscriptionType],
```
Gudahtt, line 168: For the `logs`  subscription type, this ends up being _very_ noisy. Perhaps we could add a "topic" filter to reduce the volume?

PR #206, opened 2023-01-27
v5.3.0

adonesky1:
### Added
- Add a Sign In With Ethereum (SIWE) section to test signing messages that conform to [EIP-4361 spec](https://eips.ethereum.org/EIPS/eip-4361) ([#164](https://github.com/MetaMask/test-dapp/pull/164))


Gudahtt (approved):
LGTM!

PR #207, opened 2023-02-02
Adds a Multisig Contract

seaona:
**Context**: there's been an issue with Sending ETH to Multisig Wallets, resulting in failing the transaction with the error `Cannot convert string to buffer`. See more here: https://github.com/MetaMask/metamask-extension/issues/17488

**Explanation**: 
This PR introduces a new Smart Contract for a Multisig Wallet, facilitating the work for testing this functionality on MM e2e.
The Multisig contract is the one created by Solidity By Example and can be found here.
https://solidity-by-example.org/app/multi-sig-wallet/

The subsequent work is finished on MetaMask Extension repo, with this PR:
https://github.com/MetaMask/metamask-extension/pull/17565


jpuri (approved):
Hey @seaona : changes look good. Content is huge, it might be better to create separate file.

pedronfigueiredo (approved):

PeterYinusa (approved):
LGTM!

Can we also add a button to the test dapp to deploy the contract? Can be in a separate PR

seaona:
Thank you! I will create another PR for adding the button on the UI @PeterYinusa

PR #209, opened 2023-02-06
Adds Multisig deploy and Send to Multisig functions on the frontend

seaona:
**Context**: on a previous PR, we added the Multisig Wallet bytecode and ABI. See: https://github.com/MetaMask/test-dapp/pull/207

**What**: here we add the frontend functionality for deploying a Multisig contract and Sending some ETH to its address

https://user-images.githubusercontent.com/54408225/216962360-66ce8c0a-c6b5-4951-974d-95d310347053.mp4

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,8 @@ import {
   collectiblesBytecode,
   failingContractAbi,
   failingContractBytecode,
+  multisigAbi,
+  multisigBytecode
 } from './constants.json';
 
 let ethersProvider;
@@ -25,10 +27,12 @@ let hstFactory;
 let piggybankFactory;
 let collectiblesFactory;
 let failingContractFactory;
+let multisigFactory;
 let hstContract;
 let piggybankContract;
 let collectiblesContract;
 let failingContract;
+let multisigContract;
 
 const currentUrl = new URL(window.location.href);
 const forwarderOrigin =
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -501,6 +526,55 @@ const initialize = async () => {
...
+
+    sendMultisigButton.onclick = async () => {
+      try {
+        const result = await ethereum.request({
+          method: 'eth_sendTransaction',
+          params: [
+            {
+              from: accounts[0],
+              to: multisigContract.address,
+              value: '0x16345785D8A0', // 24414062500000
```
PeterYinusa, line 563: This looks great!
Small suggestion: should we update the value to `0x0`?
See: [#129](https://github.com/MetaMask/test-dapp/pull/129)

jpuri (approved):

src/index.js
seaona: Thanks for pointing this @PeterYinusa . In this case, we are not sending ETH to an existing random or hacked address but to the contract address that we just deployed on the previous step.
Notice the same pattern for Deposit.

image

I opted for adding a small value, as I think testing with a 0 Send value might make that we miss some issues. So I really wanted to make sure that we send some value. 

What do you think?

PeterYinusa (approved):

PR #219, opened 2023-03-15
Adding an input for ERC20 token decimals

seaona:
**Context**: several issues have been reported related to sending ERC20 tokens with 0 decimals or certain value of decimals.
- https://github.com/MetaMask/metamask-extension/issues/17839
- https://github.com/MetaMask/metamask-extension/issues/18139

**What**: this PR adds the ability to input the desired amount of decimals for the ERC20 token (between 0 and 18). This way we will be able to manually test ERC20 tokens with certain problematic decimal values, like 0.

image

**How**:
- It has been observed that, that we used to pass a value `4` for the decimals in the contract constructor. However this value does not really have an effect on the token contract - it's always 4 decimals, no matter if we change it to another value. This means that the contract must have hard-coded this value (unfortunately we don't have the code for this contract). For this reason, we will need to update the ERC20 Contract, in order to make effective the value passed on the constructor. This is done in a [separate PR.](https://github.com/MetaMask/test-dapp/pull/220)

https://user-images.githubusercontent.com/54408225/225268036-6044818a-9821-4c26-b685-0be97e6cf264.mp4

- For not impacting on any of the existing tests relying on the previous value of 4 decimals, the default value will be set to 4.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -256,6 +256,18 @@
                   Token: <span id="tokenAddress"></span>
                 </p>
 
+                <div class="form-group">
+                  <label>Token Decimals</label>
+                  <input
+                    class="form-control"
+                    type="number"
+                    id="tokenDecimals"
+                    value="4"
+                    max="18"
+                    min="0"
+                  />
+                </div>
+
                 <button
                   class="btn btn-primary btn-lg btn-block mb-3"
                   id="createToken"
--- a/src/index.js
+++ b/src/index.js
@@ -116,7 +116,7 @@ const sendButton = document.getElementById('sendButton');
 const sendEIP1559Button = document.getElementById('sendEIP1559Button');
 
 // Send Tokens Section
-const decimalUnits = 4;
+const decimalUnitsInput = document.getElementById('tokenDecimals');
 const tokenSymbol = 'TST';
 const tokenAddress = document.getElementById('tokenAddress');
 const createToken = document.getElementById('createToken');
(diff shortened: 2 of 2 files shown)


PeterYinusa (changes requested):
Hmmm, I just tried it and it seems the token decimals are always `4` even if I change the input.
I wonder if the value is hardcoded in the contract 🤔

PeterYinusa:
> Hmmm, I just tried it and it seems the token decimals are always `4` even if I change the input. I wonder if the value is hardcoded in the contract 🤔

Here I entered 18 and 16, but both show as 4 
https://goerli.etherscan.io/token/0x999a0a1c8d051e190a15f4a14b5a20d4b9e6515b
https://goerli.etherscan.io/token/0x149a865bc493b0c28e4f1b8e5d5661b1270a263c

seaona:
yes @PeterYinusa , this is due to the contract that is currently at the codebase. See the **How** explanation. At the moment, the ERC20 contract that is there does not allow decimals, even if we pass them explicitly (check metamask e2e where we pass value 4, but even if you change it is always 4). 
It's happening already on main. I created [a separate PR](https://github.com/MetaMask/test-dapp/pull/220/files) for fixing this here, which introduces some other enhancements as well

PeterYinusa:
Oops, just read the rest of the PR description. Seems like this is a known issue

PeterYinusa (approved):

PR #220, opened 2023-03-15
Update ERC20 Token with Decimals & Permit support

seaona:
**Context**:  the old ERC20 token contract didn't allow for changing the decimals, when passed on the contract deployment. This has been showcasted [here](https://github.com/MetaMask/test-dapp/pull/219) even if it looks like we pass a value, it always defaults to 4. This problematic, as we need the ability to set the decimals in order to test different cases (i.e. tokens with 0 decimals).

Furthermore, having an ERC20 with Permit support, will allow us to test the Permit signature flow end-to-end. The previous Permit signature work has already been done [here](https://github.com/MetaMask/test-dapp/pull/217).

Lastly, since we didn't have the source contract, we could not actually see any of these missing properties, until it has been manually checked. So, having the contract source is convenient.

**What**: this PR updates the ERC20 token contract with 3 improvements:
- Allowing to input decimals on the constructor, so `4`decimals are not hardcoded anymore
- Adding Permit support. This will allow us to test end-to-end the Permit Signature, by grabbing the `r`, `s` and `v` values. Introduced already [here](https://github.com/MetaMask/test-dapp/pull/217)
- Adding the contract source on the `contracts` folder

**How**: the ERC20 with Permit support has been created using the [Open Zeppelin wizard](https://wizard.openzeppelin.com/), to make sure we are following the standards.
The decimal support on the constructor has been modified on the ERC20 contract base from open-zeppelin.

image


PeterYinusa (approved):
LGTM!

PR #223, opened 2023-04-11
Dependencies update

seaona:
**Context**: when installing the npm packages using the Python version `3.11` , the command fails with the error https://github.com/nodejs/node-gyp/issues/2219

Screenshot from 2023-04-11 10-13-58

**What**: this PR updates the dependencies in order to fix the error.
1. Update `"@lavamoat/allow-scripts": "^2.0.3",` using the same version from `metamask-extension`
2. Update configuration from packages using the command `yarn allow-scripts auto`

[test-dapp.webm](https://user-images.githubusercontent.com/54408225/231100363-075afe64-79bd-4d39-bfee-d15335e0c057.webm)

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "homepage": "https://metamask.github.io/test-dapp",
   "dependencies": {},
   "devDependencies": {
-    "@lavamoat/allow-scripts": "^1.0.6",
+    "@lavamoat/allow-scripts": "^2.0.3",
     "@metamask/auto-changelog": "^2.5.0",
     "@metamask/eslint-config": "^6.0.0",
     "@metamask/eslint-config-nodejs": "^6.0.0",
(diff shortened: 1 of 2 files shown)


ritave (changes requested):
Please update `.nvmrc` file to `v14` as well

ritave:
Actually, just checked [metamask-extension's `.nvmrc`](https://github.com/MetaMask/metamask-extension/blob/develop/.nvmrc), and it's set to v16, which is the minimal version we officially support for our packages.

Might as well upgrade to that one.

.github/workflows/build-lint-test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x]
```
Gudahtt, line 14: If we're updating the minimum supported Node.js version, we should update the version of Node.js referenced in the README, in `package.json`, and in `.nvmrc` as well.

seaona: aw great catch! I've updated them @Gudahtt thank you!

Gudahtt:
I have resolved just the Node.js update in #225 to unblock further publishing

PR #227, opened 2023-04-28
Add page for making a request with query parameters

Gudahtt:
We now have a page for making requests directly with query string parameters. For example, you can make an `eth_getLogs` request by navigation to:

`[test dapp path]/request.html?method=eth_getLogs&params=[{}]`

This can be useful for testing RPC methods not covered by the test dapp functionality, or for e2e testing (especially on mobile, where clicking elements on the page can be challenging).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,12 @@ module.exports = {
         sourceType: 'module',
       },
     },
+    {
+      files: ['src/request.js'],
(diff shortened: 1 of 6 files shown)


Gudahtt:
@legobeat Thanks for the review! I've just added minimal README documentation, I didn't feel good about merging this with no docs. I forgot them when initially submitting this.

PR #229, opened 2023-05-01
Fix changelog formatting

Gudahtt:
The changelog was formatted incorrectly in a few different ways:
* Various links to GitHub were mistakenly prefixed with `git+` because that's the style of link used in the `repository` field in `package.json`. The `auto-changelog` tool isn't compatible with that format, so it has been updated to use an HTTPS link. All changelog links have been fixed as well.
* The link for v6 was missing.
* The "Unreleased" section was missing.


Gudahtt:
@SocketSecurity ignore unbox-primitive@1.0.1

This report is erroneous; no dependencies are affected by this PR

PR #233, opened 2023-05-30
return all accounts from `eth_accounts`

jiexi:
* Returns all accounts from `eth_accounts` instead of only the first 

[Ticket: mme-18515](https://github.com/MetaMask/metamask-extension/issues/18515)


adonesky1 (approved):
LGTM!

PR #237, opened 2023-06-13
Collectibles -> Nfts

adonesky1:
In the extension we have dropped use of the term `collectible` in favor of `nft`. This PR simply updates the test-dapp to unify our terminology.

This will require updating our e2e tests to look grab the new text.


src/index.js
```diff
@@ -665,74 +663,74 @@ const initialize = async () => {
...
+                },
+              };
+            },
+          ),
+        );
+      } catch (error) {
+        console.error(error);
+      }
+      console.log(watchNftsResult);
+    };
```
adonesky1, line 714: This change is slightly unrelated. It is simply moving this onclick handler outside of the `mintButton` onclick handler so that it can be used to watch already minted NFTs when you pass a contract address to the dapp via url query string (without calling a new mint in that session).

PeterYinusa (approved):

digiwand (approved):

PR #241, opened 2023-06-23
Fix NFT tokenId type

adonesky1:
`tokenId` param for `wallet_watchAsset` should be type 'string'. 

Related issue: https://github.com/MetaMask/MetaMask-planning/issues/853
Related PR: https://github.com/MetaMask/metamask-extension/pull/19738


Gudahtt (approved):
LGTM!

PR #248, opened 2023-07-31
feat(): added eth address validation on contract address

mmenoxx:


src/index.js
jiexi, line 1: I just realized maybe we should do this check in a single place
instead of
```
const urlSearchParams = new URLSearchParams(window.location.search);
const deployedContractAddress = urlSearchParams.get('contract');
const scrollTo = urlSearchParams.get('scrollTo');
```

maybe?
```
const urlSearchParams = new URLSearchParams(window.location.search);
const deployedContractAddress = 
  ethers.utils.isAddress(urlSearchParams.get('contract')) 
  ? urlSearchParams.get('contract') : ''
const scrollTo = urlSearchParams.get('scrollTo');
```

mmenoxx: Good spot! Modif sent

jiexi: thank you 🙏

jiexi (approved):
Looks great. Thank you!

PR #256, opened 2023-09-07
Fix Network ID incorrectly using chainId value

jiexi:
Previously, the `chainChanged` event was being used as the network ID value. This is incorrect, as chain ID and network ID values may not match and are not guaranteed to be interchangeable. This PR fixes this issue by:
* using the `chainChanged` event to update both chainId and networkId 
* listening to the `networkChanged` event to update both chainId and networkId
  *  note that `networkChanged` is deprecated, but there is no other way to **guarantee** we don't miss a network ID change

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2093,21 +2093,9 @@ const initialize = async () => {
     ethereum.autoRefreshOnNetworkChange = false;
     getNetworkAndChainId();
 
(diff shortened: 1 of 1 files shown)


danjm (approved):
LGTM

shanejonas (approved):

PR #258, opened 2023-09-27
feat: Add PPOM setApprovalForAll button. Add token symbols for PPOM buttons

digiwand:
This PR:
- Adds Malicious Set Approval For All button
- Adds token symbols to PPOM buttons

note: We are not adding Malicious ERC721 Transfer because the user would need to hold an asset of the specific punk collection in order for the simulation to detect potential loses.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -640,10 +640,24 @@
                 </button>
                 <button
                   class="btn btn-primary btn-lg btn-block mb-3"
+                  id="maliciousERC721Transfer"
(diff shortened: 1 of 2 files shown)


seaona (approved):

seaona:
code LGTM, and I can see it working fine locally

Screenshot from 2023-09-29 15-30-00

blackdevelopa (approved):

PR #259, opened 2023-10-04
Adds ability to create multiple tokens and watch all of them.

tmashuang:
Adds ability to create multiple tokens and watch all of them. Shouldn't effect other token methods, they will use the latest created token address for approval and transfers.

For context: https://github.com/MetaMask/MetaMask-planning/issues/1407

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -281,7 +281,7 @@
                   id="watchAssets"
                   disabled
                 >
-                  Add Tokens to Wallet
+                  Add Token(s) to Wallet
                 </button>
 
                 <button
--- a/src/index.html
+++ b/src/index.html
@@ -253,7 +253,7 @@
                 </h4>
 
                 <p class="info-text alert alert-success">
-                  Token: <span id="tokenAddress"></span>
+                  Tokens: <span id="tokenAddresses"></span>
                 </p>
 
                 <div class="form-group">
@@ -278,10 +278,10 @@
 
                 <button
                   class="btn btn-primary btn-lg btn-block mb-3"
-                  id="watchAsset"
+                  id="watchAssets"
                   disabled
                 >
-                  Add Token to Wallet
+                  Add Tokens to Wallet
                 </button>
 
                 <button
--- a/src/index.html
+++ b/src/index.html
@@ -253,7 +253,7 @@
                 </h4>
 
                 <p class="info-text alert alert-success">
-                  Tokens: <span id="tokenAddresses"></span>
+                  Token(s): <span id="tokenAddresses"></span>
                 </p>
 
                 <div class="form-group">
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -1153,28 +1153,35 @@ const initialize = async () => {
...
-        method: 'wallet_watchAsset',
-        params: {
-          type: 'ERC20',
-          options: {
-            address: hstContract.address,
-            symbol: tokenSymbol,
-            decimals: decimalUnitsInput.value,
-            image: 'https://metamask.github.io/test-dapp/metamask-fox.svg',
+    watchAssets.onclick = async () => {
+      const contractAddresses = tokenAddresses.innerHTML.trim().split(' ');
```
jiexi, line 1165: maybe we can make this comma delimited like the other fields. i don't think we need the trim()

jiexi (dismissed):
small nit. LGTM otherwise

src/index.js
tmashuang: Would that change look like on L1156 to be
```tokenAddresses.innerHTML += `${hstContract.address},`;```
and then remove the `.trim()` part here?

**Review by adonesky1 — COMMENTED** [review]
Not sure I follow how this enables us to test multiple different batched token addresses. Could you include a video of the expected UX?

**Comment by tmashuang** [comment]
Currently in the test dapp, when creating a token we create a instance of the ERC20 token and use that address watchAsset, but any subsequent token creation would override the previous one. In this pr when we create a new token we append that address to the `tokens` label which when we call `wallet_watchAsset` on all of them now. 


https://github.com/MetaMask/test-dapp/assets/13376180/bc1e60c9-d433-470f-8ee6-649fac8510de

**Review by jiexi — APPROVED** [review]

File: src/index.html [hunk 2]
```diff
@@ -278,10 +278,10 @@ <h4 class="card-title">
 
                 <button
                   class="btn btn-primary btn-lg btn-block mb-3"
-                  id="watchAsset"
+                  id="watchAssets"
                   disabled
                 >
-                  Add Token to Wallet
+                  Add Tokens to Wallet
```
adonesky1, line 284: [nit]
```suggestion
                  Add Token(s) to Wallet
```

src/index.html
```diff
@@ -253,7 +253,7 @@ <h4 class="card-title">
                 </h4>
 
                 <p class="info-text alert alert-success">
-                  Token: <span id="tokenAddress"></span>
+                  Tokens: <span id="tokenAddresses"></span>
```
adonesky1, line 256: [nit]
```suggestion
                  Token(s): <span id="tokenAddresses"></span>
```

adonesky1 (approved):

jiexi (approved):

adonesky1 (approved):

PR #264, opened 2023-10-24
BREAKING: Bump to webpack 5; use Node.js >= 18

legobeat:
- Based on #223 rebased on `main`
- Update `webpack` from 4 to 5
- Add support for Node.js 18,20
- Remove support for Node.js 16, which is EOL

`webpack start` seems to run fine locally for me but reviewers are encouraged to test manually.

PR #270, opened 2023-11-03
Upgrade Node and Webpack

rekmarks:
Migrate Node from v16 (EOL) to v18 and migrate Webpack from v4 to v5.

To test, build using `yarn start` or `yarn build` and check that the output in `dist/` is as expected (no change from `main`).


package.json
```diff
@@ -67,5 +67,6 @@
       "keccak": true,
       "fsevents": true
     }
-  }
+  },
+  "packageManager": "yarn@1.22.19"
```
rekmarks, line 71: This is to ensure that correct Yarn version is used for people with Yarn >=2 installed. Can be changed when we migrate this to Yarn 3/4.

package.json
```diff
@@ -54,10 +53,11 @@
     "ethereumjs-util": "^5.1.1",
     "ethers": "^5.0.32",
     "gh-pages": "^3.1.0",
+    "node-polyfill-webpack-plugin": "^2.0.1",
```
rekmarks, line 56: Webpack 4 had Node polyfills by default. For Webpack 5 you need to do it via a plugin.

rekmarks:
Closing in favor of #264.

PR #274, opened 2023-11-21
Update `eth_signTypedData_v4` example

Gudahtt:
The example `eth_signTypedData_v4` message has been updated to include a field that can reproduce this bug: https://github.com/MetaMask/metamask-mobile/issues/7792 That signing bug results in invalid signatures for any messages that include a `bytes` field with the value `0x`.


mcmire (approved):
Looks good.

PR #275, opened 2023-11-21
Fix allow-scripts integration

Gudahtt:
The `@lavamoat/allow-scripts` integration was not setup correctly.

* The `@lavamoat/preinstall-always-fail` package was missing, despite being referenced in the configuration
* Various packages had their install scripts allowed, but they were not necessary and they failed when run.
* The `allow-scripts` step was not run in CI

All of these issues have been resolved.


mcmire (approved):
LGTM.

PR #276, opened 2023-11-21
Fix signing

Gudahtt:
The signing examples were broken as of the update to Webpack v5 (#264). The "verification" button for the signatures relies upon the Node.js `assert` and `stream` APIs. Polyfills have been added for both of these.

Additionally, the `assert` polyfill itself relies upon the Node.js `process` API, and expects `process` to be a global. A polyfill has been added for `process`, and it has been set as a global.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "@metamask/eslint-config-nodejs": "^6.0.0",
     "@metamask/onboarding": "^1.0.0",
     "@openzeppelin/contracts": "4.9.3",
+    "assert": "^2.1.0",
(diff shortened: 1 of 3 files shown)


Gudahtt:
@SocketSecurity ignore assert@2.1.0
@SocketSecurity ignore for-each@0.3.3

This author is known and trusted

Gudahtt:
@SocketSecurity ignore define-properties@1.2.1

This warning appears to be an error. The warning says that this version does not exist, but it does.

mcmire (approved):
Looks good!

PR #279, opened 2023-11-29
fix: make malicious signatures chain agnostic

seaona:
## Context
We will add support to more chains in the ppom feature [here](https://github.com/MetaMask/metamask-extension/pull/22070/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519). For that, we need a way to test the malicious signatures in different chains.
Currently is not possible, as the malicious signature data is hardcoding the chain id 1 (Mainnet)

## What
This PR changes the malicious signature data, by updating the chainId dynamically with the current chainId. This allows us to trigger the malicious signatures in different chains.

**Note**: the Malicious Trade Order is only flagged on Mainnet, and we will need to update this signature data once Blockaid provides us with another signature valid for all chains. This will come in a future PR.

## Screenshots
- Localhost (this branch version): malicious signatures working on different chains
- Test Dapp (current version): malicious signatures not working on different chains. We get the error that test dapp expects `chain id 1`

https://github.com/MetaMask/test-dapp/assets/54408225/c79a0836-17bb-4f99-bea7-432e5d49f893

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -488,8 +488,12 @@ const handleNewAccounts = (newAccounts) => {
   maxPriorityDiv.style.display = 'none';
 };
 
+let chainIdInt;
+
 const handleNewChain = (chainId) => {
   chainIdDiv.innerHTML = chainId;
+  const networkId = parseInt(networkDiv.innerHTML, 10);
+  chainIdInt = parseInt(chainIdDiv.innerHTML, 16) || networkId;
 
   if (chainId === '0x1') {
     warningDiv.classList.remove('warning-invisible');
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -1970,17 +1974,14 @@ const initializeFormElements = () => {
       }
     } catch (err) {
       console.error(err);
-      signTypedDataV3VerifyResult.innerHTML = `Error: ${err.message}`;
+      signTypedDataVerifyResult.innerHTML = `Error: ${err.message}`;
```
seaona, line 1977: this is an unrelated fixed, but encountered the issue along the way: we were using the singTypedDataV3 field instead of the correct typed data version

src/index.js
```diff
@@ -1970,17 +1974,14 @@ const initializeFormElements = () => {
...
     }
   };
 
   /**
    * Sign Typed Data Version 3 Test
    */
   signTypedDataV3.onclick = async () => {
-    const networkId = parseInt(networkDiv.innerHTML, 10);
-    const chainId = parseInt(chainIdDiv.innerHTML, 16) || networkId;
-
```
seaona, line 1983: after factorizing the code, we can remove all these blocks from all the signatures, as we now have a variable holding the chain id integer

jpuri (approved):

DDDDDanica (approved):

PR #280, opened 2023-11-30
fix: remove suggested gas from malicious transactions

seaona:
## Description
This is a small fix that removes the suggested gas values from the malicious transactions. This increases the chance for the transactions to be flagged as malicious, as the gas set was a bit high, making the validation fail with `not enough funds` error, if the account don't have enough ETH

## Screenshots
Before (test-dapp)
After (localhost)

https://github.com/MetaMask/test-dapp/assets/54408225/9af807ee-ada6-4aa6-abb2-ffbee475953e


seaona:
closing in favor of https://github.com/MetaMask/test-dapp/pull/281

PR #281, opened 2023-11-30
add: support malicious tx's in multiple networks

seaona:
## Description
The PPOM module now is supported in more networks. For that reason, we need to update the transactions in order to point to the correct contract addresses according to the network we have selected.
The Trade Order signature also needs to be changed, as the previous one was not flagged in all networks.
Notice for seeing the Malicious warning, you need to have some balance, for paying the gas for the transactions, and a small amount of the ERC20 token for the Transfer transaction, on each chain.

Pending (in progress)
- [x] handle mainnet (2 diff ERC20)
- [x] handle other not supported networks

## Screenshots

https://github.com/MetaMask/test-dapp/assets/54408225/5d6b45aa-5a71-4eb4-b00a-7be6c0618ac0

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1309,12 +1309,19 @@ const initializeFormElements = () => {
 
   // Malicious ERC20 Approval
   maliciousApprovalButton.onclick = async () => {
+    let erc20Contract;
+
+    if (networkName) {
+      erc20Contract = ERC20_SAMPLE_CONTRACTS[networkName];
+    } else {
+      erc20Contract = '0x4fabb145d64652a948d72533023f6e7a623c7c53';
+    }
     const result = await ethereum.request({
       method: 'eth_sendTransaction',
       params: [
         {
           from: accounts[0],
-          to: `${ERC20_SAMPLE_CONTRACTS[networkName]}`,
+          to: erc20Contract,
           gas: '0x30d40',
           data: '0x095ea7b3000000000000000000000000e50a2dbc466d01a34c3e8b7e8e45fce4f7da39e6000000000000000000000000000000000000000000000000ffffffffffffffff',
           gasPrice: '0x76c3b0342',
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -1307,7 +1314,7 @@ const initializeFormElements = () => {
       params: [
         {
           from: accounts[0],
-          to: '0x4fabb145d64652a948d72533023f6e7a623c7c53',
+          to: `${ERC20_SAMPLE_CONTRACTS[networkName]}`,
```
seaona, line 1317: we change dynamically the contract address depending on the network we are in, in order for the transaction to be flagged as malicious, from the sample data Blockaid provided

src/index.js
```diff
@@ -1367,7 +1374,7 @@ const initializeFormElements = () => {
...
..."erc20Token"},{"type":"uint256","name":"erc20TokenAmount"},{"type":"Fee[]","name":"fees"},{"type":"address","name":"erc721Token"},{"type":"uint256","name":"erc721TokenId"},{"type":"Property[]","name":"erc721TokenProperties"}],"Fee":[{"type":"address","name":"recipient"},{"type":"uint256","name":"amount"},{"type":"bytes","name":"feeData"}],"Property":[{"type":"address","name":"propertyValidator"},{"type":"bytes","name":"propertyData"}],"EIP712Domain":[{"name":"name","type":"string"},{"name":"version","type":"string"},{"name":"chainId","type":"uint256"},{"name":"verifyingContract","type":"address"}]},"domain":{"name":"ZeroEx","version":"1.0.0","chainId":${chainIdInt},"verifyingContract":"0xdef1c0ded9bec7f1a1670819833240f027b25eff"},"primaryType":"ERC721Order","message":{"direction":"0","maker":"${accounts[0]}","taker":"0x5FbDB2315678afecb367f032d93F642f64180aa3","expiry":"2524604400","nonce":"100131415900000000000000000000000000000083840314483690155566137712510085002484","erc20Token":"0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2","erc20TokenAmount":"42000000000000","fees":[],"erc721Token":"0x8a90CAb2b38dba80c64b7734e58Ee1dB38B8992e","erc721TokenId":"2516","erc721TokenProperties":[]}}`,
```
seaona, line 1377: malicious trade order signature was not being flag on the new networks Blockaid support, so we change this signature data so, this new signature is flagged in all networks

src/index.js
```diff
@@ -1302,15 +1309,20 @@ const initializeFormElements = () => {
...
     const result = await ethereum.request({
       method: 'eth_sendTransaction',
       params: [
         {
           from: accounts[0],
-          to: '0x4fabb145d64652a948d72533023f6e7a623c7c53',
-          gas: '0x30d40',
+          to: erc20Contract,
           data: '0x095ea7b3000000000000000000000000e50a2dbc466d01a34c3e8b7e8e45fce4f7da39e6000000000000000000000000000000000000000000000000ffffffffffffffff',
-          gasPrice: '0x76c3b0342',
```
seaona, line 1313: we remove suggested gas to avoid gas issues, so gas is calculated based on the network we are - we don't need to change this for each network

PR #298, opened 2024-03-05
feat: add malicious tx and signatures for sepolia

seaona:
## Description
This PR adds support for testing malicious transactions and signatures on Sepolia.
Note: for the ERC20 transfer, if you do not own any tokens, the validation will fail. For testing this specific case, I've added a Mint button above (only for Sepolia) which allows you to mint tokens. After minting, trying the ERC20 malicious transfer should successfully validate that as malicious.

## Screenshots

https://github.com/MetaMask/test-dapp/assets/54408225/84970827-1203-491e-8c8e-00e905d648ba

## Manual Testing
Testing malicious tx and signatures:
1. Select Sepolia
2. Try Malicious Send Eth
3. Mint some ERC2 tokens
4. Try Malicious Send ERC
5. Try Malicious Approval
6. Try Set Approval for all
7. Try all the signatures

Testing mint button
1. Select Sepolia and refresh --> see Mint button is there
2. Select another network and refresh --> see Mint button is not there

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -270,6 +270,7 @@ const addEthereumChain = document.getElementById('addEthereumChain');
 const switchEthereumChain = document.getElementById('switchEthereumChain');
 
 // PPOM
+const mintERC20 = document.getElementById('mintERC20');
 const maliciousApprovalButton = document.getElementById(
   'maliciousApprovalButton',
 );
@@ -361,6 +362,7 @@ const allConnectedButtons = [
   maliciousPermit,
   maliciousTradeOrder,
   maliciousSeaport,
+  mintERC20,
 ];
 
 // Buttons that are available after initially connecting an account
@@ -403,6 +405,7 @@ const initialConnectedButtons = [
   maliciousPermit,
   maliciousTradeOrder,
   maliciousSeaport,
+  mintERC20,
 ];
 
 // Buttons that are available after connecting via Wallet Connect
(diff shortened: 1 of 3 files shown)


segun (approved):

src/index.js
```diff
@@ -1478,6 +1487,24 @@ const initializeFormElements = () => {
    *  PPOM
    */
 
+  // Mint ERC20 in Sepolia
+  mintERC20.onclick = async () => {
```
cryptotavares, line 1491: should the `mintERC20` be renamed to provide more context? Currently the button only displays if we are in Sepolia, and the erc20 contract is set to a the [Sepolia one](https://github.com/MetaMask/test-dapp/pull/298/files#diff-972434016582e4f749542bddfc5a1473440bbdcdd8f32ddfdb5f3bfd239a2f4bR17). So `mintSepoliaERC20` seems to be more appropriate.
What do you think?

src/index.js
```diff
@@ -655,6 +659,11 @@ const handleNewChain = (chainId) => {
 
 const handleNewNetwork = (networkId) => {
   networkDiv.innerHTML = networkId;
+  if (networkId === ('11155111' || '0xaa36a7')) {
```
cryptotavares, line 662: Can we make this more readable? Was wondering if we could leverage the `NETWORKS_BY_CHAIN_ID` const.. but might not be worth the effort. We could do something simpler like:
```js
const isNetworkIdSepolia = networkId === ('11155111' || '0xaa36a7');

if (isNetworkIdSepolia) {
  ...
}
```

PR #304, opened 2024-03-18
Adds transferFromTokens error to token addresses text box

pedronfigueiredo:
As part of [#2224](https://github.com/MetaMask/MetaMask-planning/issues/2224), we need to have a way to test whether the token allowance to transfer tokens is exceeded or not. This change shows the error message on the token addresses text box so that it can be asserted against in e2e tests.

Screenshot 2024-03-18 at 16 52 59

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1767,15 +1767,19 @@ const initializeFormElements = () => {
   };
 
   transferFromTokens.onclick = async () => {
-    const result = await hstContract.transferFrom(
-      transferFromSenderInput.value,
-      transferFromRecipientInput.value,
-      decimalUnitsInput.value === '0'
-        ? 1
-        : `${1.5 * 10 ** decimalUnitsInput.value}`,
-      { from: accounts[0] },
-    );
-    console.log('result', result);
+    try {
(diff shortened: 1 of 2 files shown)


seaona:
Overall looks good. Just a nit: do you think we could maybe add a new field below the actions, to display the error there?
This html item is currently missing for the ERC20 section, and if we add it, it would make it more similar to what's in other contracts and we could advocate for this pattern in the future too.

Screenshot from 2024-03-19 09-43-45

Feel free to ignore this though, and it could also be done in another PR

seaona (approved):

src/index.js
```diff
@@ -1767,15 +1770,20 @@ const initializeFormElements = () => {
...
+    try {
+      const result = await hstContract.transferFrom(
+        transferFromSenderInput.value,
+        transferFromRecipientInput.value,
+        decimalUnitsInput.value === '0'
+          ? 1
+          : `${1.5 * 10 ** decimalUnitsInput.value}`,
+        { from: accounts[0] },
+      );
+      console.log('result', result);
```
jpuri, line 1782:
```suggestion
```

PR #306, opened 2024-03-22
fix: Fix network status field

Gudahtt:
The network status field would show the `networkId` on load (in decimal), but after switching chains it would show the `chainId` (in hex). It now shows the `networkId` in all cases.

This means re-introducing the deprecated `networkChanged` event listener, but this listener is still something we want to be able to test.


tommasini (approved):
LGTM!

PR #310, opened 2024-03-26
Add Base network support.

sleepytanya:
Add Base chain support for Blockaid.


seaona (approved):
Nice!! I see that both contracts work as expected

Screenshot from 2024-03-27 17-02-39

Screenshot from 2024-03-27 17-02-48

PR #313, opened 2024-04-08
feat: add deeplinks for Send Eth, Send ERC20 and Approve ERC20

seaona:
## Description
This PR adds support to Deeplinks in the test dapp. This will allows us to automate e2e tests around the feature. Specifically,
- Adds Simple Send Eth for deeplinks
- Adds ERC20 transfer. For this, we need to pass the contract address and the decimals in the URL. Notice for this functionality we'll access it from the browser (outside the wallet), that's why we need to pre-deploy a contract, in order to use it, if we want to have assets.
- Adds ERC20 approve. For this, we need to pass the contract address and the decimals in the URL. Notice for this functionality we'll access it from the browser (outside the wallet), that's why we need to pre-deploy a contract, in order to use it, if we want to have assets.

## Screenshots

Screenshot from 2024-04-08 20-03-33

https://github.com/MetaMask/test-dapp/assets/54408225/723e133e-9cb6-4422-b4c1-764e1c632c08

## Manual QA
1. Build the test dapp locally
2. With the Extension/Mobile test dapp: deploy an ERC20 token
3. Copy the address the the decimals (default is 4)
4. Now, go to the Mobile Browser (outside the MetaMask wallet)
5. Add this URL: `http://localhost:9011/?contract=0x35e22143f85CD31F5319Ae4295B36F396bF27f78&decimals=4` but change the contract address and decimals for the one you deployed
6. Click Send Eth --open MM and proceed with the tx
7. Go back to the browser
8. Click Transfer Tokens -- open MM and proceed with the tx
9. Go back to the browser
10. Click Approve Tokens -- open MM and proceed with the tx

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,6 +46,11 @@ if (!ethers.utils.isAddress(deployedContractAddress)) {
   deployedContractAddress = '';
 }
 
+let tokenDecimals = urlSearchParams.get('decimals');
+if (!tokenDecimals) {
+  tokenDecimals = '18';
+}
+
 const scrollTo = urlSearchParams.get('scrollTo');
 
 /**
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -46,6 +46,11 @@ if (!ethers.utils.isAddress(deployedContractAddress)) {
   deployedContractAddress = '';
 }
 
+let tokenDecimals = urlSearchParams.get('decimals');
+if (!tokenDecimals) {
+  tokenDecimals = '18';
```
seaona, line 51: we set this value as default, if no value for `decimals` is passed in the query param on the URL

digiwand (approved):

PR #284, opened 2023-12-12
feat: add Wallet Connect  support

seaona:
## Context
Needed for testing the PPOM module in Mobile + Wallet Connect

## What
Adds wallet connect support to the test dapp.

## How
- A button triggers the Wallet Connect Modal. Using the Wallet Connect option, you can scan the QR code using your phone based MetaMask Wallet, to start getting transaction/signatures requests triggered from the test dapp on your phone wallet.
- Whenever the WalletConnect connection is established, a new provider is added. You can then choose to use MetaMask or WalletConnect providers alternatively to trigger your Extension or your Mobile wallet
- For Wallet Connect, we only have a subset of functionalities enabled for now: Sends/Signatures but excluding Contract Deployments, as needs some more investigation on how to get the contract address. This can be extended in the future (not currently needed for the purposes of PPOM).
- When we use the `?contract` URI param on the URL, all the buttons (except deploy contracts) are also enabled for Wallet Connect to use, in the same was as it's done for Extension

## Screenshots

https://github.com/MetaMask/test-dapp/assets/54408225/57fda92e-6e76-40b5-9f7c-48c81ee03db3

Screenshot from 2023-12-11 16-19-41

## Future Work
- Refactor button list and in general the index.js file, as it starts to grow big. We could better organize the different dapp functionalities in the future
- Add Deploy Smart Contract functionality for Wallet Connect
- Add SDK support in the same way -- this will enable to test PPOM with SDK

## Manual Testing Steps
1. Build the test dapp locally `yarn` and `yarn start`
2. Go to the browser

### Testing the Connection
3. Click Wallet Connect and select Wallet Connect option
4. With your phone, scan the QR code within MM wallet
    - See connection prompts to the phone
5. Accept connect
    - See Wallet Connect Provider is added on top of the test-dapp
    - See current selected provider is now Wallet Connect
    - See Wallet Connect buttons are enabled (Sends and signatures)

### Testing tx/signatures with WC
6. Trigger any tx/signature
    - See the tx/signature is triggered on your phone

### Testing Network switch handling
7. Switch networks
    -  See status network on the test dapp is updated

### Testing switching providers back and forth (Extension-Mobile)
8. Click Use Metamask Main
    - See current selected provider is now MetaMask Main
    - See Extension Connected buttons are all enabled
9. Trigger another tx/signature
    - See now the Extension wallet is triggered
10. Click Use Wallet Connect
    - See current selected provider is back to Wallet Connect

### Testing the old functionality
[...]

PR #331, opened 2024-05-02
feat: add sdk support

seaona:
### Description
This PR adds support for SDK Connection as well as moves the logic for WalletConnect + SDKConnect to its own file.

For the future, we could also move the provider logic and updatingButtons etc, to a utils file, since those functions are quite heavy and the index.js file is growing too big.
Leaving this out of scope of this PR as would be harder to review.

### Screenshots

https://github.com/MetaMask/test-dapp/assets/54408225/98f85f38-33d1-4ab5-a141-43d9d5d6b95c

### Manual QA
Check that WalletConnect continues to work (any tx or signature)
Check that SDK connect works as expected (any tx or signature)


src/index.html
```diff
@@ -122,7 +122,7 @@ <h3 class="card-title">
             <div class="card">
               <div class="card-body">
                 <h4 class="card-title">
-                  Basic Actions
+                  Connect Actions
```
seaona, line 125: I found that this name might be more explicit

src/index.html
```diff
@@ -152,11 +159,6 @@ <h4 class="card-title">
               </div>
             </div>
           </div>
-        </div>
-      </section>
-
-      <section>
-        <div class="row d-flex justify-content-center">
```
seaona, line 159: I found this improves a little bit the alignment, by not taking so much space: now Connect Actions and BAsic Actions will be in the same line

Screenshot from 2024-05-03 12-39-10


before:

Screenshot from 2024-05-03 12-40-11

src/index.js
```diff
@@ -494,49 +500,6 @@ const initialConnectedButtons = [
   maliciousPermitIntAddress,
 ];
 
-// Buttons that are available after connecting via Wallet Connect
```
seaona, line 497: since this is now fixed, we can remove this section:

> For Wallet Connect, we only have a subset of functionalities enabled for now: Sends/Signatures but excluding Contract Deployments, as needs some more investigation on how to get the contract address. This can be extended in the future (not currently needed for the purposes of PPOM).

[see ref](https://github.com/MetaMask/test-dapp/pull/284)

src/index.js
```diff
@@ -548,62 +511,33 @@ let scrollToHandled = false;
...
+    );
+  });
 };
 
-const modal = createWeb3Modal({
-  ethersConfig: defaultConfig({ metadata }),
-  projectId,
-});
-
-async function handleWalletConnectChange({ isConnected }) {
```
seaona, line 570: all this logic is now moved in the connections.js file

andreahaku (approved):
LGTM

src/index.js
```diff
@@ -8,13 +7,18 @@ import {
...
+  walletConnect,
+} from './connections';
 import Constants from './constants.json';
 import {
-  NETWORKS_BY_CHAIN_ID,
   ERC20_SAMPLE_CONTRACTS,
   ERC721_SAMPLE_CONTRACTS,
+  NETWORKS_BY_CHAIN_ID,
 } from './onchain-sample-contracts';
+import { getPermissionsDisplayString, stringifiableToHex } from './utils';
```
seaona, line 21: taking the chance to re-order the imports by alphabetical order

src/index.js
```diff
@@ -994,14 +942,7 @@ const updateFormElements = () => {
     }
     clearDisplayElements();
   }
-  if (
-    isWalletConnectConnected &&
-    activeProviderNameResult.innerText === 'wallet-connect'
-  ) {
-    for (const button of walletConnectButtons) {
```
seaona, line 1001: we don't have wallet connect specific buttons anymore since the previous issue was fixed, so we can remove this logic

PR #354, opened 2024-08-13
[MMSDK] Updates mmsdk to latest and fixes a bad call when extension doesn't exist

christopherferreira9:
Previously upon checking if the extension is present via EIP6963 the dapp would perform `eth_accounts` and sync chainId but when using the SDK this conflicts as a connection might not exist at that point. This avoids that empty call if extension is not installed and the SDK is used.


andreahaku (approved):
LGTM

PR #356, opened 2024-08-22
feat: Add malicious contract interaction as an option

pnarayanaswamy:
Adds a button Malicious Contract Interaction to the PPOM section 

image

image

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -39,6 +39,10 @@ const {
  * Page
  */
 
+const SEPOLIA_NETWORK_ID_HEX = '0xaa36a7';
+const SEPOLIA_NETWORK_ID_DEC = '11155111';
+const BASE_NETWORK_ID = '8453';
+
 const currentUrl = new URL(window.location.href);
 const forwarderOrigin =
   currentUrl.hostname === 'localhost' ? 'http://localhost:9010' : undefined;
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -1604,6 +1630,30 @@ const initializeFormElements = () => {
     console.log(result);
   };
 
+  // Malicious Mint
```
seaona, line 1633:
```suggestion
  // Malicious Contract Interaction
```

src/index.js
```diff
@@ -1604,6 +1630,30 @@ const initializeFormElements = () => {
...
+      contractAddress = '0x00008F1149168C1D2fa1eBa1Ad3e9cD644510000';
+    }
+
+    const result = await provider.request({
+      method: 'eth_sendTransaction',
+      params: [
+        {
+          from: accounts[0],
+          to: contractAddress,
+          data: '0xef5cfb8c0000000000000000000000000b3e87a076ac4b0d1975f0f232444af6deb96c59',
```
seaona, line 1649: interesting! this calldata is calling the function `claimRewards` but that doesn't exist in the contracts I've checked. However I do see the Blockaid malicious warning, so I guess it's fine, even if that would make it revert

src/index.js
```diff
@@ -1604,6 +1630,30 @@ const initializeFormElements = () => {
...
   };
 
+  // Malicious Mint
+  maliciousContractInteractionButton.onclick = async () => {
+    let contractAddress;
+
+    if (networkName === 'mainnet') {
+      contractAddress = '0x000062Accd1a9d62eF428eC86cA3dD4f45120000';
+    } else {
+      contractAddress = '0x00008F1149168C1D2fa1eBa1Ad3e9cD644510000';
```
seaona, line 1640: nit: I would rather move this into the [sample contracts file](https://github.com/MetaMask/test-dapp/blob/main/src/onchain-sample-contracts.js#L33), to make it easier to locate and maintain in the future (but not blocking)

pnarayanaswamy: done

src/index.js
pnarayanaswamy: done

src/index.js
pnarayanaswamy: the payload and the accounts are what blockaid are using to test it, so I guess it is ok indeed

seaona (approved):
LGTM! nice addition :heart_eyes:

Blakerose8 (commented):
Ok

Blakerose8 (commented):
Ok

PR #372, opened 2024-10-31
DRAFT: Jl/1193 multichain example dapp

jiexi:
Please don't review. This is more for storage, less for review. Not intending to merge these changes into this repo.

### DIFF
--- a/src/eip1193_index.html
+++ b/src/eip1193_index.html
@@ -1,178 +1,211 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <title>Ethereum EIP-1193 API Dapp</title>
-</head>
-<body>
-  <h1>Ethereum EIP-1193 API Dapp</h1>
-  <button id="connectButton">Connect Wallet</button>
-  <br>
-  <div id="connectedAccounts">Connected Accounts: Not available</div>
-  <br><br>
-  <label for="chainSelect">Select Chain:</label>
-  <select id="chainSelect">
-    <option value="0x1">Ethereum Mainnet</option>
-    <option value="0xe708">Linea Mainnet</option>
-    <option value="0xaa36a7">Sepolia Testnet</option>
-    <option value="0xe705">Linea Sepolia</option>
-  </select>
-  <br><br>
-  <button id="sendTxButton">Send 0 Ether to Null Address</button>
-  <br><br>
-  <div id="currentChain">Current Chain: Not available</div>
-  <div id="blockHead">Current Block Head: Not available</div>
+  <head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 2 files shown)


src/multichain_demo.js
```diff
@@ -0,0 +1,1151 @@
...
+//
+// Constants
+//
+const WeiPerEth = 1000000000000000000
+
+const BridgeableScopes = {
+    // Sepolia
+    "eip155:11155111": {
+        name: "Sepolia",
+        contractAddress: '0x2e2512fd69cba059DFf557cD6f683a3279402e91',
```
jiexi, line 15: note you will have to deploy your own dummy bridge contracts to each of these testnets as the `withdraw` method is locked to my wallet addresses for the currently deployed contracts listed in this constant. The solidity file is in this PR. Note that there is no actual bridging, just piggy bank contracts that mock the bridging flow that is:
1. sending to originating chain bridge contract
2. claiming on target chain bridge contract

src/multichain_demo.js
```diff
@@ -0,0 +1,1151 @@
...
+const NODE_TEXT_START = "-2.4em";
+const NODE_TEXT_SPACING = "1.2em";
+const NODE_FLASH_COLOR = "gray"
+const EDGE_COLOR = "#999";
+const EDGE_STROKE_WIDTH = 3;
+const FLASH_DURATION = 3000;
+
+const ETH_VALUE_PRECISION = 4;
+
+const USE_SUBSCRIPTIONS = true; // use false for polling
```
jiexi, line 67: set this to false to use polling instead of eth_subscribe to trigger on chain data fetching.

funnily enough, using polling results in a more responsive dapp experience than using subscription on block heads to determine when to read the chain for new data. I think this is because eth subscriptions causes us to hit the RPC rate limit, where as just using a setTimeout poll is close enough and results in less rate limiting

src/multichain_index.html
jiexi, line 1: Screenshot 2024-11-01 at 2 03 20 PM

example dapp to illustrate what the multichain enables over the eip1193 version

src/eip1193_index.html
jiexi, line 1: Screenshot 2024-11-01 at 2 04 28 PM

simple example eip1193 dapp to illustrate chain switching friction

src/multichain_demo.html
jiexi, line 1: Screenshot 2024-11-01 at 2 05 56 PM

More complicated bridging demo dapp to highlight some of the features enabled by the Multichain API.

Note this code is garbage spaghetti nonsense because I don't know how to use d3 and relied on chatgpt to start the project which was probably a mistake. And then I just kept bandaging things on in an effort to finish this before I am OOO. So... sorry lol

src/multichain_demo.js
jiexi: BTW I used the online Remix IDE to compile and deploy. EVM version: Paris (because of Linea not supporting an opcode)

PR #379, opened 2024-12-17
Refactor: Move all components to separate files

Montoya:
This refactors everything as independent modules with shared global context. We also attempted to

- make these modules as self-contained as possible.
- move constants and functions into separate files.
- use shared constants and objects for sharing state.
- add events and listeners where appropriate for handling state changes.

This refactor does not: 

- Change the connect / disconnect / provider logic very much
- Rewrite with React

But it's progress


src/components/signatures/signTypedDataV3-sign.js
```diff
@@ -0,0 +1,176 @@
+import { recoverTypedSignature } from '@metamask/eth-sig-util';
+import { toChecksumAddress } from 'ethereumjs-util';
+import globalContext from '../..';
+import { EIP712Domain } from '../../signatures/utils';
+
+export function signTypedDataV3Component(parentContainer) {
+  parentContainer.insertAdjacentHTML(
+    'beforeend',
+    `<div class="col-xl-4 col-lg-6 col-md-12 col-sm-12 col-12 d-flex align-items-stretch">
```
pedronfigueiredo, line 9: This is great! Would it make sense to keep a separate HTML file for each component so that it has syntax highlighting, and import the markup here?

Montoya: Yes. There are other ways to do it too, like importing a JSX parser so you can use JSX fragments instead of strings.

Montoya: (There's also [tricks to making this work](https://dev.to/boyum/using-tagged-template-literals-to-syntax-highlight-code-in-strings-4h1i), or just bit the bullet and port to React :) )

pnarayanaswamy:
Ran all e2e successfully with a patch version from local: https://github.com/MetaMask/metamask-extension/pull/29433

pnarayanaswamy (dismissed):

pedronfigueiredo (approved):

seaona (changes requested):
hi there! Nice work :fire: 
I'm in the middle of the testing, will update this comment with my findings.

So far, I found that whenever I am using the Wallet Connect functionalty several buttons remain disabled (deploy contracts, signatures..) see video with the comparision of current prod and this test dapp version below:

https://github.com/user-attachments/assets/59bcab49-9905-4091-89f2-39c0bdd0ab30

src/components/signatures/signTypedData-variants.js
```diff
@@ -0,0 +1,149 @@
...
+                <p class="info-text alert alert-warning">
+                  Result:
+                  <span id="signPermitVariantResult"></span>
+                  <p class="info-text alert alert-warning" id="signPermitVariantResultR">r: </p>
+                  <p class="info-text alert alert-warning" id="signPermitVariantResultS">s: </p>
+                  <p class="info-text alert alert-warning" id="signPermitVariantResultV">v: </p>
+                </p>
+              </div>
+            </div>
+          </div>`,
```
seaona, line 59: Nice new section! Suggestion: could we add a `Verify` section too, so we are sure the wallet works correctly with those signature values?
No need to be in this PR, but would be nice to have this, at some point

Montoya:
> hi there! Nice work 🔥 I'm in the middle of the testing, will update this comment with my findings.
> 
> So far, I found that whenever I am using the Wallet Connect functionalty several buttons remain disabled (deploy contracts, signatures..) see video with the comparision of current prod and this test dapp version below:
> 
>  wallet-connect-btn-disabled.mp4 
> [Update] The rest seems to work fine on my end!

Please try again, just merged a fix by @pnarayanaswamy

seaona (approved):
thanks, I see the issue fixed now!!

https://github.com/user-attachments/assets/c16c54c0-3670-4cab-afd9-130d1d81bf48

src/components/transactions/eip747.js
seaona, line 1: nit: this functionality seems to be incorrectly placed under transactions

src/sample-networks.js
seaona, line 1: nit: this file could be consoldiated with the network data we already have here

https://github.com/MetaMask/test-dapp/blob/main/src/onchain-sample-contracts.js#L2

src/components/ppom/transactions.js
```diff
@@ -0,0 +1,307 @@
...
+          >
+            Malicious Trade Order
+          </button>
+          <button
+            class="btn btn-primary btn-lg btn-block mb-3"
+            id="maliciousSeaport"
+            disabled
+          >
+            Malicious Seaport
+          </button>
```
seaona, line 87: nit: this section is placed under ppom/transactions but those are actually signatures

src/components/signatures/malformed-transactions.js
seaona, line 1: the following content is related to transactions, but it's placed under signatures folder

seaona:
The functionality seems to be preserved so I've approved the PR, but recently added a couple of nits regarding the re-org, non-blocking but might be worth taking a look now or in a future PR

src/components/signatures/malformed-transactions.js
Montoya: Yes, the way the cards are grouped is based on the name of the section. It's not ideal.

src/components/ppom/transactions.js
Montoya: Yes, this is how the test-dapp was originally organized. We could rename transactions.js to something else.

Montoya:
@seaona I think overall the IA nits should be for the developers who made these cards. I don't want to mess with how the cards are grouped and sorted, my intention was to keep the IA the same.

src/components/signatures/malformed-transactions.js
seaona: About the malformed tx/signatures, we made them already as separated cards

Captura desde 2025-01-22 08-49-38

src/components/ppom/transactions.js
seaona: so in this case, when we built this, we created 2 subsections one for Transactions and one for Signatures. They were in the same card, but If that's problematic with the current organization we could have one card for TX and a separate one for Signatures. What do you think?

Captura desde 2025-01-22 08-51-29

seaona:
hi there :wave: thanks for the reply!! I made a couple of the cards mentioned above so I'm happy to add the changes if they make sense to you!

src/components/signatures/malformed-transactions.js
seaona: mm yeah so I think something could be improved there, as at the moment the current folder naming doesn't completely match the classification :thinking: 




mm if I imagine if I want to add a new section, let's say Solana tx's  and signatures, where will they go?
I think the natural way might be to create a solana folder, and inside, to have a signature component and then a tx component.

So I think we might want to follow the same approach here. We could have a folder for Malformed Interactions and inside we can distinguish between  signatures and tx. Same goes for Blockaid and so on.

-- Malformed Interactions
 | 
 |--- Signatures
 |--- Transactions


-- Malicious Interactions
 | 
 |--- Signatures
 |--- Transactions

-- Solana
 | 
 |--- Signatures
 |--- Transactions

What do you think?

Captura desde 2025-01-22 09-33-54

Captura desde 2025-01-22 09-34-11

PR #388, opened 2025-02-11
feat: add EIP-5792 section

matthewwalsh0:
## Description

Add an EIP-5792 section including the following buttons:

- `Send Calls`
- `Get Calls Status`
- `Get Capabilities`

Also support custom `wallet_sendCalls` requests via the `Edit Request` button and dynamic form.

Note that including `capabilities` in `wallet_sendCalls` is not yet supported as that would likely be specific to each capability once supported by MetaMask.

## Screenshots

Section

Get Calls Status

Get Capabilities

Edit Request


OGPoyraz (approved):

src/components/transactions/eip5792/sendCalls.js
```diff
@@ -0,0 +1,254 @@
+import globalContext from '../../..';
+
+const DEFAULT_CALLS = [
+  {
+    to: '0x0c54FcCd2e384b4BB6f2E405Bf5Cbc15a017AaFb',
+    data: '0x654365436543',
+    value: '0x1234123412341234',
+  },
+  {
+    to: '0x5FbDB2315678afecb367f032d93F642f64180aa3',
```
seaona, line 10: Q: that's a malicious address, is it intended to have a malicious send in the batch?

matthewwalsh0: No, I was just looking for another checksummed address to test with, I'll use a different one to avoid confusion 😄

src/components/transactions/eip5792/sendCalls.js
```diff
@@ -0,0 +1,254 @@
+import globalContext from '../../..';
+
+const DEFAULT_CALLS = [
+  {
+    to: '0x0c54FcCd2e384b4BB6f2E405Bf5Cbc15a017AaFb',
+    data: '0x654365436543',
+    value: '0x1234123412341234',
```
seaona, line 7: mmm that's `4.837 ETH`, for the e2e is fine, but in the case we want to manually use the test dapp that's a ton of money :laughing: could we maybe tweak the values to be either a very low amount? Enough for testing simulations, but still very few to be able to perform a couple of manual flows

Screenshot from 2025-02-11 17-15-19

src/components/transactions/eip5792/sendCalls.js
```diff
@@ -0,0 +1,254 @@
...
+  },
+  {
+    to: '0x5FbDB2315678afecb367f032d93F642f64180aa3',
+    data: '0x789078907890',
+    value: '0x4321432143214321',
```
seaona, line 12: just a note, maybe in a future PR we could add contract interactions in the batch too (like approve, transfer etc etc).
Summary of combinations that seem important:
- batch sends all tx benign
- batch send with malicious tx
- batch sends + contract interactions all benign
- batch sends + contract interactions with malicious
- batch send/contract with a failing tx

I can support on adding those if this make sense to you too :pray:

src/components/transactions/eip5792/sendCalls.js
matthewwalsh0: Good point, will lower the default, but these also can be customised via `Edit Request`.

seaona: oh yeahhh, make sense, thank you!

src/components/transactions/eip5792/sendCalls.js
matthewwalsh0: A contract interaction would be great, but harder to default as we'd have to use known addresses on each chain.

The `Edit Request` would let you test any combination manually.

I think default buttons for malicious and failing make sense in future, but the malicious one for example can't be tested yet until it's on Sepolia for example or a chain that Blockaid supports.

seaona (approved):
LGTM!!

src/components/transactions/eip5792/sendCalls.js
seaona: the Edit Request is :fire:

PR #389, opened 2025-02-11
fix: update send calls default calls

matthewwalsh0:
Update default calls in `wallet_sendCalls` request.

- Reduce `value` to 1 Gwei and 0.5 Gwei.
- Use random address rather than malicious address for second call.
- Add `version` to request.

### DIFF
--- a/src/components/transactions/eip5792/sendCalls.js
+++ b/src/components/transactions/eip5792/sendCalls.js
@@ -1,5 +1,7 @@
 import globalContext from '../../..';
 
+const VERSION = '1.0';
+
 const DEFAULT_CALLS = [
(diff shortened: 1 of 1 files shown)


seaona (approved):

src/components/transactions/eip5792/sendCalls.js
```diff
@@ -143,6 +145,7 @@ export function sendCallsComponent(parentContainer) {
     const calls = useInputs ? getCalls() : DEFAULT_CALLS;
 
     return {
+      version: VERSION,
```
seaona, line 148: nice catch!

PR #401, opened 2025-04-10
add: malicious send calls

seaona:
## Description
This PR adds a new section for malicious batch transactions.
On each send call we'll have 1 type of a malicious transaction, the rest will be the 2 default simple sends (benign).

## Screenshots

Screenshot from 2025-04-11 15-40-32

https://github.com/user-attachments/assets/b358e987-804a-4003-b69a-59e335f970b9

### DIFF
--- a/src/components/ppom/batching.js
+++ b/src/components/ppom/batching.js
@@ -159,7 +159,7 @@ export function ppomMaliciousBatchingAndQueueing(parentContainer) {
           params: [
             {
               from: globalContext.accounts[0],
-              to: '0x5FbDB2315678afecb367f032d93F642f64180aa3',
+              to: maliciousAddress,
               value: '0x0',
               gasLimit: MIN_GAS_LIMIT,
               maxFeePerGas: '0x2540be400',
--- a/src/components/ppom/sharedConstants.js
+++ b/src/components/ppom/sharedConstants.js
@@ -3,10 +3,12 @@ import {
   ERC20_SAMPLE_CONTRACTS,
   ERC721_SAMPLE_CONTRACTS,
   MALICIOUS_CONTRACT_ADDRESSES,
+  NETWORKS_BY_CHAIN_ID,
 } from '../../onchain-sample-contracts';
 
 export const getMaliciousTransactions = (globalContext) => {
-  const networkName = globalContext.networkName || 'default';
+  const chainId = globalContext.chainIdInt;
+  const networkName = NETWORKS_BY_CHAIN_ID[chainId] || 'default';
 
   return {
     eth: {
--- a/src/components/ppom/sharedConstants.js
+++ b/src/components/ppom/sharedConstants.js
@@ -6,11 +6,18 @@ import {
   NETWORKS_BY_CHAIN_ID,
 } from '../../onchain-sample-contracts';
 
+const maliciousTransactionsCache = {};
+
 export const getMaliciousTransactions = (globalContext) => {
   const chainId = globalContext.chainIdInt;
   const networkName = NETWORKS_BY_CHAIN_ID[chainId] || 'default';
 
-  return {
+  // to avoid recalculating the tx object everytime if the chainId hasn't changed
+  if (maliciousTransactionsCache[chainId]) {
+    return maliciousTransactionsCache[chainId];
+  }
+
+  const transactions = {
     eth: {
       to: maliciousAddress,
       value: '0x9184e72a000',
@@ -38,4 +45,7 @@ export const getMaliciousTransactions = (globalContext) => {
       value: '0x0',
     },
   };
+
+  maliciousTransactionsCache[chainId] = transactions;
+  return transactions;
 };
--- a/src/components/ppom/sharedConstants.js
+++ b/src/components/ppom/sharedConstants.js
@@ -1,0 +1,40 @@
+// src/components/ppom/sharedConstants.js
+import { maliciousAddress } from '../../sample-addresses';
+import {
+  ERC20_SAMPLE_CONTRACTS,
+  ERC721_SAMPLE_CONTRACTS,
+  MALICIOUS_CONTRACT_ADDRESSES,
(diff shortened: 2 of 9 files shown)


src/components/ppom/transactions.js
```diff
@@ -139,20 +136,18 @@ export function ppomMaliciousTransactionsAndSignatures(parentContainer) {
 
   document.addEventListener('newNetwork', function (e) {
     mintSepoliaERC20.hidden = !isSepoliaNetworkId(e.detail.networkId);
-    maliciousContractInteractionButton.hidden =
-      isBaseNetworkId(e.detail.networkId) ||
-      isSepoliaNetworkId(e.detail.networkId);
```
seaona, line 144: this seems a mistake, as the logic prevented to show the Contract Interaction button for Sepolia (unsure why this was added). Now we can see the button back again, and test this flow

src/components/ppom/transactions.js
```diff
@@ -228,18 +206,15 @@ export function ppomMaliciousTransactionsAndSignatures(parentContainer) {
 
   // Malicious Contract interaction
   maliciousContractInteractionButton.onclick = async () => {
-    const contractAddress =
-      MALICIOUS_CONTRACT_ADDRESSES[globalContext.networkName] ||
-      MALICIOUS_CONTRACT_ADDRESSES.default;
```
seaona, line 233: we directly use the malicious contract interaction data from the sample contracts. We don't want to hide this button (unsure why this was added)
The payload for the hex data is what is malicious, no matter to which address we point (in this case, Blockaid flags the tx as malicious because of the hex data, not checking the to address -- which makes our life easier)

src/components/ppom/sharedConstants.js
```diff
@@ -0,0 +1,39 @@
+import { maliciousAddress } from '../../sample-addresses';
```
seaona, line 1: moved the malicious tx's in this file, so they can be used both in the ppom tx and in the ppom batch txs

src/components/ppom/transactions.js
```diff
@@ -51,17 +47,17 @@ export function ppomMaliciousTransactionsAndSignatures(parentContainer) {
...
+            Malicious Set Approval For All
           </button>
           <button
             class="btn btn-primary btn-lg btn-block mb-3"
-            id="maliciousSetApprovalForAll"
+            id="maliciousContractInteractionButton"
             disabled
           >
-            Malicious Set Approval For All
+            Malicious Contract Interaction
```
seaona, line 60: re-ordered these 2 as it makes more sense like this (first all custom contract interactions, and the last one the fallback contract interaction) -- also following this order for the batch

src/onchain-sample-contracts.js
```diff
@@ -1,35 +1,38 @@
...
   arbitrum: '0x8659a4876369b94515a86048fe7f99daba6b9a7d',
-  sepolia: '0xbba60aa8144579e07c6db64121b0f608ab6f0c89',
+  avalanche: '0x880f7e04D722e305126F7E1efd3434A7d5b1465c',
   base: '0x90997fc967e75b7e69f899133aab31d197beb802',
+  bsc: '0xebfbfd7c41b123500fb16b71c43b400c12b08be0',
+  mainnet: '0xBC4CA0EdA7647A8aB7C2061c2E118A18a936f13D',
   opBnb: '0x61d7e121185b1d7902a3da7f3c8ac9faaee8863b',
+  optimism: '0xaf18644083151cf57f914cccc23c42a1892c218e',
+  polygon: '0x9e8ea82e76262e957d4cc24e04857a34b0d8f062',
+  sepolia: '0xbba60aa8144579e07c6db64121b0f608ab6f0c89',
```
seaona, line 35: reorder alphabetically

src/components/ppom/batching.js
```diff
@@ -159,7 +159,7 @@ export function ppomMaliciousBatchingAndQueueing(parentContainer) {
           params: [
             {
               from: globalContext.accounts[0],
-              to: '0x5FbDB2315678afecb367f032d93F642f64180aa3',
```
seaona, line 162: remove hardcoded malicious address, and use the const

src/sample-addresses.js
```diff
@@ -1 +1 @@
-export const maliciousAddress = '0x5FbDB2315678afecb367f032d93F642f64180aa3';
```
seaona, line 1: this address is no longer flagged as malicious (https://consensys.slack.com/archives/C08JQ6DDCSU/p1744401213662199?thread_ts=1744173737.539829&cid=C08JQ6DDCSU) replacing with a new address which isflagged

seaona: the change has been reverted, as the address i working again as malicious

src/components/ppom/sharedConstants.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+    maliciousContractInteraction: {
+      to:
+        MALICIOUS_CONTRACT_ADDRESSES[networkName] ||
+        MALICIOUS_CONTRACT_ADDRESSES.default,
+      data: '0xef5cfb8c0000000000000000000000000b3e87a076ac4b0d1975f0f232444af6deb96c59',
+      value: '0x0',
+    },
+  };
+};
```
digiwand, line 41: [comment] thinking out loud - there's probably another way to write this function to optimize it so that it doesn't create a new object for each call, but overall it's not a heavy operation and calling getMaliciousTransactions is nice and readable

digiwand (approved):

src/components/ppom/sharedConstants.js
seaona: yeah this rings a bell.. let me give it a thought, thank you for your comment @digiwand :pray:

seaona: updated! thanks for the suggestion :pray:

digiwand (approved):

PR #404, opened 2025-04-15
Upgrade workflows to use ubuntu-latest

mcmire:
The `build-lint-test` workflow uses the Ubuntu 20.04 image, which GitHub deprecated some time ago. Because of this new PRs will no longer pass CI. This commit upgrades the workflow to `ubuntu-latest` which should always work (and matches what we do in other repos).


seaona (approved):
thank you

PR #405, opened 2025-04-16
add: malicious send call bypasses

seaona:


seaona:
just a test branch, deleting now

PR #414, opened 2025-08-19
feat: add network picker

seaona:
## Description
This PR adds a network picker to be able to switch networks from the dapp.

- If the network is already in the wallet, and we gave permissions to that network, it switches networks.
- If the network is already in the wallet, but we didn't give permissions to that network, it triggers the MM permissions dialog and then the network is switched
- If the network is not present in the wallet, we throw an error in the UI.

Future work:
- support addEtehreumChain for non-added networks, so we can add that network and then switch to it 

## Screenshots

Main networks

Screenshot from 2025-08-19 11-38-32

Test networks

Screenshot from 2025-08-19 11-38-35

Happy Path

https://github.com/user-attachments/assets/ce85fb5c-71ff-4817-96be-8ae75316fa27

Error Handling

https://github.com/user-attachments/assets/ac7b7c74-32b3-4735-872c-3f907f36be5f

#### Flow 1: Connected + Permissions
1. Connect to the test dapp
2. Give permissions to a cople of networks
3. Click on Network picker
4. See modal with network list appears
5. Select another network that you currently have on the Wallet and you gave permissions to
6. See network is switch and chain id is switch
7. Trigger a tx/signature, see your updated network

#### Flow 2: Connected without Permissions
1. Connect to the test dapp
2. Give permissions to just 1 network
3. Click on Network picker
4. See modal with network list appears
5. Select another network that you currently have on the Wallet and you didn't give permissions to
6. See MM dialog appears with the Permissions screen
7. Accept
8. See network is switch and chain id is switch
9. Trigger a tx/signature, see your updated network

#### Flow 3: Connected without Network in the Wallet
1. Click on the Network picker
2. Select a network that you don't currently have on the wallet
3. See error on the UI and nothing happens

### DIFF
--- a/src/components/connections/networks-helpers.js
+++ b/src/components/connections/networks-helpers.js
@@ -125,11 +125,45 @@ export function createNetworkItem(network) {
   `;
 
   item.addEventListener('click', async () => {
+    hideNetworkError(); // Hide any existing error before attempting to switch
     await switchNetwork(network.chainId);
     document.querySelector('.network-modal').style.display = 'none';
   });
 
   return item;
+}
+
+export function showNetworkError(message) {
+  // Remove any existing error message
+  hideNetworkError();
+
+  // Create error message element
+  const errorDiv = document.createElement('div');
+  errorDiv.id = 'networkError';
+  errorDiv.className = 'error-message';
+  errorDiv.style.marginTop = '10px';
+  errorDiv.style.width = '100%';
+  errorDiv.innerHTML = `
+    <img src="alert-red.svg" alt="" class="error-message-icon" />
+    <div class="error-message-text">${message}</div>
+  `;
+
+  // Find the network picker button and insert error after it
+  const networkButton = document.getElementById('openNetworkPicker');
+  const cardBody = networkButton.closest('.card-body');
+  cardBody.appendChild(errorDiv);
+
+  // Auto-hide after 5 seconds
+  setTimeout(() => {
+    hideNetworkError();
+  }, 5000);
+}
+
+export function hideNetworkError() {
+  const existingError = document.getElementById('networkError');
+  if (existingError) {
+    existingError.remove();
+  }
 }
 
 export async function switchNetwork(chainId) {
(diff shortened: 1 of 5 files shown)


src/components/connections/networks-helpers.js
```diff
@@ -0,0 +1,232 @@
...
+  errorDiv.style.width = '100%';
+  errorDiv.innerHTML = `
+    <img src="alert-red.svg" alt="" class="error-message-icon" />
+    <div class="error-message-text">${message}</div>
+  `;
+
+  // Find the network picker button and insert error after it
+  const networkButton = document.getElementById('openNetworkPicker');
+  const cardBody = networkButton.closest('.card-body');
+  cardBody.appendChild(errorDiv);
```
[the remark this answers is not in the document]
seaona: I've removed the icon alert-red.svg
Since networksComponent is called during the main app initialization and creates the entire structure, these elements should always be present. The defensive checks are probably unnecessary.

src/components/connections/networks.js
```diff
@@ -0,0 +1,74 @@
...
+        </div>
+      </div>
+    </div>
+  `;
+  document.body.appendChild(modal);
+
+  const openButton = document.getElementById('openNetworkPicker');
+  const closeButton = modal.querySelector('.network-modal-close');
+
+  // Populate network lists
```
ffmcgee725, line 51: Method name is pretty self descriptive 👍🏾
so I'd prefer we remove this comment to avoid redundancy

src/components/connections/networks-helpers.js
```diff
@@ -0,0 +1,229 @@
...
+  }
+}
+
+export function updateCurrentNetworkDisplay() {
+  const currentNetworkName = document.getElementById('currentNetworkName');
+
+  if (
+    globalContext.chainIdInt !== undefined &&
+    globalContext.chainIdInt !== null
+  ) {
```
ffmcgee725, line 196:
```suggestion
if (globalContext.chainIdInt) {
```
from what I understand, we don't care what type of value this is, as long as it's defined, so
this condition can be simplified

src/components/connections/networks.js
seaona: great point, I've removed all of them :pray:

src/components/connections/networks-helpers.js
```diff
@@ -0,0 +1,229 @@
...
+      currentNetworkName.textContent = `Current Network: ${network.name}`;
+    } else {
+      // Fallback to chain ID if network not found
+      currentNetworkName.textContent = `Current Network: Chain ID 0x${globalContext.chainIdInt.toString(
+        16,
+      )}`;
+    }
+  } else {
+    currentNetworkName.textContent = 'Current Network: Not Connected';
+  }
```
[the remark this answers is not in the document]
seaona: whoa much better :heart_eyes: Added the changes :pray: thank you vmuch!

Thatguysdino (commented):
****

Thatguysdino:
We the people, thank you

